PR #3, opened 2018-01-19
telemetry poller example application from apollo fusion demo [WIP]

jacoffey3:


jacoffey3:
[Example Application](https://trello.com/c/pKu5ryi7/170-example-application)

applications/example-telemetry-poller/main.py
```diff
@@ -0,0 +1,62 @@
...
+    subsystem,
+    param,
+    value,
+    timestamp
+  }
+}
+"""
+
+headers = {'Content-Type' : 'application/graphql'}
+PAYLOAD_DEST = "http://{}:{}".format(cfg['APP_IP'], cfg['PAYLOAD_PORT'])
```
ChristopherRabotin: Don't mind me much since I'm just browsing, here's an unsolicited tip: Python's `format` is perfect for dictionaries since you can simply specify the dictionary as the argument, and retrieve different keys within the string itself (albeit while making the code slightly confusing).
```
In [1]: d = {'ip': '127.0.0.1', 'port': 1590}

In [2]: "http://{0[ip]}:{0[port]}".format(d)
Out[2]: 'http://127.0.0.1:1590'
```

plauche: @ChristopherRabotin Thanks for the tip! I didn't realize you could do that!

PR #4, opened 2018-01-19
Adding blurbs about partuuid process and IDs

CatGarab:
PRs kubos/kubos-linux-build#83 and kubos/uboot#22 change the way that the BBB and MBM2 boot up. I was going to change the relevant documentation, but apparently that was never documented...

So instead I'm adding a new blurb about how U-Boot gets the boot device for Linux and then the new disk signatures that are used to generate the partuuids.


CatGarab:
[BBB - handle only eMMC](https://trello.com/c/QarGCPuH/147-bbb-handle-only-emmc)

docs/linux-docs/kubos-linux-on-mbm2.rst
```diff
@@ -45,6 +45,10 @@ U-Boot
 This board utilizes U-Boot's SPL feature. A small boot file called "MLO" is
 run and that file then loads the main U-Boot image into SDRAM.
 
+The main U-Boot image iterates through the `boot_targets` variable to attempt 
+to boot from an available MMC device. The partuuid of the first successful
```
plauche: How much of this documentation could be consolidated instead of duplicated across both the bbb and mbm2 pages? Something to think about for a future docs push.

docs/linux-docs/kubos-linux-on-mbm2.rst
```diff
@@ -45,6 +45,10 @@ U-Boot
...
+device is passed off to Linux to be used to mount the root filesystem. 
+
 KubOS Linux Build Process
```
plauche: nit: KubOS -> Kubos

docs/linux-docs/kubos-linux-on-bbb.rst
```diff
@@ -45,6 +45,10 @@ U-Boot
 This board utilizes U-Boot's SPL feature. A small boot file called "MLO" is
 run and that file then loads the main U-Boot image into SDRAM.
 
+The main U-Boot image iterates through the `boot_targets` variable to attempt 
+to boot from an available MMC device. The partuuid of the first successful
+device is passed off to Linux to be used to mount the root filesystem. 
+
 KubOS Linux Build Process
```
plauche: nit: KubOS -> Kubos

docs/linux-docs/kubos-linux-on-mbm2.rst
CatGarab: A fair amount. These docs (also the scripts and things in KLB) are generally *okay* for now, but could definitely use a makeover. What I would really love is some magical tool that allows a user to select their board, and then these docs (building/upgrading/working with {target}) would be automatically assembled based upon that target's valid sections. That way the users would literally only see the information relevant to them.

PR #22, opened 2018-01-25
Automating docs deployment

plauche:
* Created script to generate docs from within circleci container
* Create script to separately upload them
* Updated circleci container with docs tools
* Created new docker container for handling docs generation


tools/dist/.dockerignore
```diff
@@ -0,0 +1 @@
+!.git/
```
plauche: By default our docker container ignores this directory, which makes it very difficult to detect the current git tag to use in the docs...

prep_docs.sh
```diff
@@ -0,0 +1,13 @@
...
+
+echo "Tag for docs: $latest_tag"
+
+# Place current docs tag into docs conf
+sed -i "s/0.0.0/$latest_tag/g" docs/conf.py
+
+echo "Generating new docs"
+./tools/gendocs.py --version $latest_tag
+
+chmod a+rw html/
```
plauche: Not totally sure why but after running this script inside of Docker this folder can't be modified. Specifically in the "native" Circle environment.

docs/conf.py
```diff
@@ -55,9 +55,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = u'1.1.0'
+version = u'0.0.0'
 # The full version, including alpha/beta/rc tags.
-release = u'1.1.0'
+release = u'0.0.0'
```
plauche: We now inject the version as the docs are generated.

circle.yml
```diff
@@ -17,3 +17,5 @@ deployment:
     branch: master
     commands:
       - ./deploy.sh
+      - docker run -t -v $PWD:$PWD --workdir $PWD kubos/kubos-dev:latest sh prep_docs.sh
```
plauche: The creation and deployment have been split into two separate scripts. Creation is done inside of the Docker container where we have the necessary docs tools. Deployment is done in the "native" Circle environment where we have the necessary ssh keys to access the web server.

prep_docs.sh
kyleparrott: docker runs as the `root` user in containers- possibly permissions error if files created by root user?

deploy_docs.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+latest_tag=`git tag --sort=-creatordate | head -n 1`
+
+echo "Uploading new docs.."
+git clone https://github.com/kubos/pm-tools
+./pm-tools/docs/upload_docs.sh $latest_tag ./html/
```
kyleparrott: does this set the landing redirect from `docs.kubos.co` to `http://docs.kubos.co/<this tag>/index.html` ? or does the server just default to the highest tag?

Do we want the default docs landing to land on build releases? or only update the default docs users see on major releases?

plauche: That is correct. I don't really have an issue with `docs.kubos.co` always reflecting the latest docs, as master is supposed to always be releasable. @jacoffey3 @CatGarab @davesims thoughts?

jacoffey3: I think it's okay to have it always reflect master. It'll make us take more concern to put real effort into the docs we push. We'll be doing doc reviews for releases anyways.

kyleparrott: > as master is supposed to always be releasable

I agree with that. But Vagrant ships with a frozen major release, not the latest master build tag at the time of users cloning.

I have a bit of concern on the following case:
- I clone the latest vagrant environment
- I go to the docs website. The default docs I see may or may not match the default source in my environment.
  - I don't see notice or way to change the version of docs I'm looking at.

You guys can overrule my concern - but as a user of the framework - if the api docs were out of date w/ my environment by default and it resulted in me wasting any amount of time I would be upset.

kyleparrott: To be clear - I don't have an issue with pushing docs on every merge, I think that's a good idea. 

My complaint is that I believe docs.kubos.co should land on the latest major release by default. If you want to go look at the latest master build docs, there should be an option through the UI to do so. 

Just like the CLI only showing major releases by default (there's an option to show build releases too). Our docs experience should mirror the beliefs/opinions in our tools.

CatGarab: Not quite sure how I didn't notice this conversation before. I agree with Kyle. Most users will be using the current release of our vagrant box, so the docs should reflect that by having a major release be the default available docs, with the master version available elsewhere

docs/Dockerfile
```diff
@@ -0,0 +1,22 @@
+# kubos/kubos-docs:0.1.0
+
+FROM phusion/baseimage:0.9.22
```
plauche: @kyleparrott I decided to split the docs stuff off into it's own docker container separate from the `kubos-dev` one.

plauche:
This PR is ready for a re-review. New behavior, when combined with this [commit](https://github.com/kubos/pm-tools/commit/dc828571fdd2406e2f41696af508389e62548d7f), is to automatically upload docs after every passing Circle build. Tags which include a build version will be uploaded to http://docs.kubos.co/master/, tags without a build version will be considered a "release" and uploaded to http://docs.kubos.co/.

Yes pm-tools should have been reviewed, whoops.

tools/gendocs.py
```diff
@@ -45,6 +45,8 @@ def main():
 
     subprocess.call("sphinx-build docs/ html/", shell=True)
 
+    shutil.rmtree("./xml")
```
CatGarab: Why are you removing this directory?

plauche: This is a temp directory holding the XML that Doxygen generates.

deploy_docs.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+latest_tag=`git tag --sort=-creatordate | head -n 1`
+
+echo "Uploading new docs.."
```
CatGarab: nit: ellipses have three periods

deploy_docs.sh
```diff
@@ -0,0 +1,7 @@
...
+git clone https://github.com/kubos/pm-tools
```
CatGarab: This is a private repo. Is any kind of account information/authorization needed?

tools/gendocs.py
CatGarab: Yeah, but Circle is already creating a temporary VM. It's all going to get deleted when it's done, right?

plauche:
@CatGarab the build version uploading is done in a script in the `pm-tools` repo https://github.com/kubos/pm-tools/commit/dc828571fdd2406e2f41696af508389e62548d7f.

tools/gendocs.py
plauche: Yup but if you ran this locally to generate the docs then you'd have the dangling xml folder.

PR #23, opened 2018-01-25
Adding Rust wrapper for kubos-hal-iobc

plauche:
- Create FFI interface to C-based supervisor functions
- Create proper Rust interfaces to wrap around FFI functions
- Tests to ensure valid data conversion

### DIFF
--- a/hal/kubos-hal-iobc/src/ffi.rs
+++ b/hal/kubos-hal-iobc/src/ffi.rs
@@ -1,0 +1,41 @@
+//
+// Copyright (C) 2017 Kubos Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use libc::uint8_t;
+
+pub const LENGTH_TELEMETRY_HOUSEKEEPING: usize = 37;
+pub const LENGTH_TELEMETRY_GET_VERSION: usize = 34;
+pub const LENGTH_COMPILE_INFORMATION: usize = 19;
+pub const SUPERVISOR_NUMBER_OF_ADC_CHANNELS: usize = 10;
+
+/// There isn't a big need to replicate the C-union/struct representation
+/// of the data only to have it moved into a Rust struct. So we will just
+/// use an array of uint8_t to capture the data and later move it into
+/// a proper Rust struct.
+#[repr(C)]
+pub struct supervisor_version(pub [uint8_t; LENGTH_TELEMETRY_GET_VERSION]);
+
+#[repr(C)]
+pub struct supervisor_housekeeping(pub [uint8_t; LENGTH_TELEMETRY_HOUSEKEEPING]);
+
+/// Bring in C functions from kubos-hal-iobc
+extern "C" {
+    pub fn supervisor_emergency_reset() -> bool;
+    pub fn supervisor_reset() -> bool;
+    pub fn supervisor_powercycle() -> bool;
+    pub fn supervisor_get_version(version: *mut supervisor_version) -> bool;
+    pub fn supervisor_get_housekeeping(housekeeping: *mut supervisor_housekeeping) -> bool;
+}
--- a/hal/kubos-hal-iobc/src/lib.rs
+++ b/hal/kubos-hal-iobc/src/lib.rs
@@ -1,0 +1,338 @@
+//
+// Copyright (C) 2017 Kubos Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+extern crate libc;
+
+use std::mem;
+
+mod ffi;
+
+/// Structure returned by supervisor_version
+#[derive(Debug)]
+pub struct SupervisorVersion {
+    pub dummy: u8,
+    pub spi_command_status: u8,
+    pub index_of_subsystem: u8,
+    pub major_version: u8,
+    pub minor_version: u8,
+    pub patch_version: u8,
+    pub git_head_version: u32,
+    pub serial_number: u16,
+    pub compile_information: Vec<i8>,
+    pub clock_speed: u8,
+    pub code_type: i8,
+    pub crc: u8,
+}
+
+/// Structure for individual enable statuses
+/// Used in SupervisorHousekeeping
+#[derive(Debug)]
+pub struct SupervisorEnableStatus {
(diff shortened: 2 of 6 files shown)


plauche:
[iOBC Supervisor Rust Wrapper](https://trello.com/c/zZvBxJ7L/137-iobc-supervisor-rust-wrapper)

hal/kubos-hal-iobc/src/ffi.rs
```diff
@@ -0,0 +1,41 @@
...
+/// There isn't a big need to replicate the C-union/struct representation
+/// of the data only to have it moved into a Rust struct. So we will just
+/// use an array of uint8_t to capture the data and later move it into
+/// a proper Rust struct.
+#[repr(C)]
+pub struct supervisor_version(pub [uint8_t; LENGTH_TELEMETRY_GET_VERSION]);
+
+#[repr(C)]
+pub struct supervisor_housekeeping(pub [uint8_t; LENGTH_TELEMETRY_HOUSEKEEPING]);
+
```
CatGarab: why isn't the `enable_status` structure also here?

hal/kubos-hal-iobc/src/ffi.rs
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+//
+
+use libc::uint8_t;
+
+pub const LENGTH_TELEMETRY_HOUSEKEEPING: usize = 37;
+pub const LENGTH_TELEMETRY_GET_VERSION: usize = 34;
+pub const LENGTH_COMPILE_INFORMATION: usize = 19;
+pub const SUPERVISOR_NUMBER_OF_ADC_CHANNELS: usize = 10;
+
```
CatGarab: Some of the other `LENGTH_` defines have been left out. Is that on purpose?

hal/kubos-hal-iobc/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+pub struct SupervisorVersion {
+    pub dummy: u8,
+    pub spi_command_status: u8,
+    pub index_of_subsystem: u8,
+    pub major_version: u8,
+    pub minor_version: u8,
+    pub patch_version: u8,
+    pub git_head_version: u32,
+    pub serial_number: u16,
+    pub compile_information: Vec<i8>,
```
CatGarab: This should probably be an array, since it's a fixed number of elements (`LENGTH_COMPILE_INFORMATION`)

hal/kubos-hal-iobc/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+/// Structure returned by supervisor_housekeeping
+#[derive(Debug)]
+pub struct SupervisorHousekeeping {
+    pub dummy: u8,
+    pub spi_command_status: u8,
+    pub enable_status: SupervisorEnableStatus,
+    pub supervisor_uptime: u32,
+    pub iobc_uptime: u32,
+    pub iobc_reset_count: u32,
+    pub adc_data: Vec<u16>,
```
CatGarab: Should also probably be an array

hal/kubos-hal-iobc/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+        Ok(convert_raw_version(version))
+    }
+}
+
+/// Converts raw bytes from iOBC into SupervisorHousekeeping
+fn convert_raw_housekeeping(raw: ffi::supervisor_housekeeping) -> SupervisorHousekeeping {
+    SupervisorHousekeeping {
+        dummy: raw.0[0] as u8,
+        spi_command_status: raw.0[1] as u8,
+        enable_status: SupervisorEnableStatus {
```
CatGarab: `supervisor_enable_status_t` has an extra `raw_value` byte that you might need to handle/skip

hal/kubos-hal-iobc/src/lib.rs
```diff
@@ -0,0 +1,338 @@
...
+            (raw.0[11] as u32) | (raw.0[12] as u32) << 8 | (raw.0[13] as u32) << 16
+                | (raw.0[14] as u32) << 24
+        },
+        adc_data: {
+            // combining bytes into 16-bit uints
+            let mut v = Vec::<u16>::new();
+            for i in 0..(ffi::SUPERVISOR_NUMBER_OF_ADC_CHANNELS) {
+                v.push((raw.0[15 + 2 * i] as u16) | (raw.0[15 + 2 * i + 1] as u16) << 8);
+            }
+            v
```
CatGarab: extra character?

hal/kubos-hal-iobc/src/lib.rs
davesims: Great catch @CatGarab ☝️ Is it possible that value would ever change or be made variable?

hal/kubos-hal-iobc/src/ffi.rs
plauche: These structures exist solely to wrap arrays used by `supervisor_get_version` and `supervisor_get_housekeeping` to fetch data. The `enable_status` structure isn't explicitly used by either of these functions, the data is inside of the array filled in by `supervisor_get_housekeeping`.

hal/kubos-hal-iobc/src/ffi.rs
plauche: Yup, I only pulled over the defines needed for interfacing with the data that `supervisor_get_version` and `supervisor_get_housekeeping` return back. The rest of those defines in `supervisor.h` are used internally in `supervisor.c` as part of the supervisor comms.

hal/kubos-hal-iobc/src/lib.rs
plauche: Hmm I think I have it accounted for. The whole of `enable_status` only spans a single byte. There are two blank bits between `supervisor_mode` and `busy_rtc` which should be the difference between `( & 0x04 ) >> 2` and `( & 0x20 ) >> 5`. There is also another blank bit after `power_off_rtc`, but I don't believe anything needs to be done with it.

hal/kubos-hal-iobc/src/lib.rs
plauche: The single `v` is shorthand for `return v`.

hal/kubos-hal-iobc/src/lib.rs
CatGarab: Within that byte, the bit masks you have set up look correct.
However, looking at the original structure, it still looks like 2 bytes to me.
```
typedef union __attribute__((__packed__)) {
    /** Raw value of the version configuration bytes. */
    uint8_t raw_value;  <---------------------------- Byte 1
    /** Individual enable status fields */
    struct __attribute__((__packed__)) supervisor_enable_status_fields_t
    {
                
        uint8_t <------------------------------------ Byte 2
                /** OBC Power. */
                power_obc : 1, 
                /** Output power to the RTC. */
                power_rtc : 1,
                /** Supervisor mode. */
                is_in_supervisor_mode : 1, 
                : 1, : 1,
                /** RTC is busy. */
                busy_rtc : 1,
                /** RTC is doing power off. */
                power_off_rtc : 1,
                : 1;
    } /** Individual enable status fields */ fields;
} supervisor_enable_status_t;
```

plauche: The `supervisor_enable_status_t` is a union of a single `uint8_t` and the struct `supervisor_enable_status_fields_t`. Inside of that struct we have a single `uint8_t` which is then split into several bit fields.  Because of the union those two bytes are the same byte.

hal/kubos-hal-iobc/src/lib.rs
CatGarab: Cool, that makes sense. I'd like it if the variable name was a little longer so that it doesn't look like just a random character the next time I haven't had coffee yet :smile:

hal/kubos-hal-iobc/src/lib.rs
CatGarab: Ahhhhh, I totally missed that it said "union". My bad. :+1:

PR #26, opened 2018-01-31
EyeStar-D2 Duplex API

plauche:
### Implement basic commanding for the Near Space Launch D2 Duplex Radio

Implement core communications API, Connection, Radio, Commanding and such, with an eye to reuse as the basic interface for KubOS SDK applications.

- [x] Basic Traits: Radio, Connection, Message
- [x] D2-specific implementations: D2Duplex, D2Message, etc.
- [x] Uploaded file command, file count & related structs
- [x] state_of_health_record command & related structs
- [x] num_as_bytes crate
- [x] Shore up error handling
/cc @jacoffey3


davesims:
[Duplex D2](https://trello.com/c/oYxARbC8/141-duplex-d2)

radio/nsl-duplex-d2/src/duplex_d2.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub fn get_state_of_health_record(&self) -> Result<StateOfHealthRecord, String> {
+        match self.send_command(GET_MODEM_STATE_OF_HEALTH) {
+            Ok(response) => Ok(StateOfHealthRecord::new(response)),
+            Err(e) => Err(e),
+        }
+    }
+
+    fn send_command(&self, command: u64) -> Result<Vec<u8>, String> {
+        match self.conn.send(command.as_bytes()) {
+            Ok(_) => match self.conn.receive() {
```
davesims: Timeout?

radio/nsl-duplex-d2/src/file.rs
```diff
@@ -1,20 +1,37 @@
...
+            .ok_or("Invalid buffer length")?;
+
+        Ok(data.to_vec())
+    }
 
-        let name = String::from_utf8(file_response[11..(11 + name_size)].to_vec()).unwrap();
-        let payload = file_response[(11 + name_size)..(11 + name_size + payload_size)].to_vec();
+    pub fn from_response(file_response: &[u8]) -> Result<File, String> {
+        let name_size = size_from_utf_8(file_response[2..5].to_vec());
+        let data_size = size_from_utf_8(file_response[5..11].to_vec());
```
creationix: I don't think this `size_from_utf_8` function is a good idea.  It has two places it can panic and in rust it's discouraged from storing indexes or sizes into slices/vectors/strings as separate variables because then the lifetimes are decoupled.

creationix: See listing 4.6 here https://doc.rust-lang.org/book/second-edition/ch04-03-slices.html

creationix: Ok, I see, the input data contains length headers as ascii numbers it appears.  is that right?  Sorry I didn't understand the goal here.

creationix: What do you think of this non panic version of `size_from_utf_8`?

```rust
fn size_from_utf_8(utf8_size: &[u8]) -> Result<usize, String> {
    Ok(str::from_utf8(utf8_size)
        .or(Err("utf8 parse error"))?
        .parse()
        .or(Err("decimal parse error"))?)
}
```

You can use it like the following:

```rust
let name_size = size_from_utf_8(&file_response[2..5])?;
let data_size = size_from_utf_8(&file_response[5..11)?;
```

We never create strings, just work with slices directly which I believe never copies the data.

creationix: I do wonder about using the `.or(Err("some string"))` as a way to convert the error to a string.

creationix: If we create our own error type with from conversions for the error types we care to support when it can be automated!

creationix: A [simple way](https://rustbyexample.com/error/multiple_error_types/boxing_errors.html) to make error handling easy is to just use `Box<std::error::Error>` as the error type in the result.  

Also I realized that manually checking the bounds and using `[]` access on slices is a lot easier than converting to vectors and messing with the `.get()` method.

Using that I've simplified the entire `from_response` to a single, fairly simple function.

```rust
use std::str;                                                                   
use std::error;                                                                  
                                                                                 
pub struct File {                                                                
    pub name: String,                                                            
    pub data: Vec<u8>,                                                           
}                                                                                
                                                                                 
impl File {                                                                      
    pub fn from_response(data: &[u8]) -> Result<File, Box<error::Error>> {       
        if data.len() < 11 {                                                     
            return Err("Not enough data in response".into());                    
        }                                                                        
                                                                                 
        let name_size: usize = str::from_utf8(&data[2..5])?.parse()?;            
        let data_size: usize = str::from_utf8(&data[5..11])?.parse()?;           
        let name_start = 11;                                                     
        let name_end = name_start + name_size;                                   
        let data_start = name_end;                                               
        let data_end = data_start + data_size;                                   
                                                                                 
        if data.len() < data_end {                                               
            return Err("Not enough data in response".into());                    
        }                                                                        
                                                                                 
        let name = str::from_utf8(&data[name_start..name_end])?.to_string();     
                                                                                 
        let data = (&data[data_start..data_end]).to_vec();                       
                                                                                 
        Ok(File { name, data })                                                  
    }                                                                            
}                                                                                
```

creationix:
@davesims that's fine as long as we go back at some point and fix the places where invalid network data can cause panics. Namely the `[]` access on slices without length checks and the `.unwrap()` calls.

davesims:
@creationix actually I'd prefer fixing all of that in the code review process here before we merge. Mind annotating those in a review?

radio/nsl-duplex-d2/src/file.rs
creationix: If we don't want to change the error type yet, we can add back in the `.or(Err("some string"))` in front of each `?`.  It's a little verbose, but will catch the errors and route them without panicking.

creationix: Try this version:
```rust
use std::str;                                                                    
                                                                                 
pub struct File {                                                                
    pub name: String,                                                            
    pub data: Vec<u8>,                                                           
}                                                                                
                                                                                 
impl File {                                                                      
    pub fn from_response(data: &[u8]) -> Result<File, String> {                 
        if data.len() < 11 {                                                     
            return Err("Not enough data in response".to_string());              
        }                                                                        
                                                                                 
        let name_size: usize = str::from_utf8(&data[2..5])                      
            .or(Err("Invalid utf8 data"))?                                       
            .parse()                                                             
            .or(Err("Invalid decimal"))?;                                        
        let data_size: usize = str::from_utf8(&data[5..11])                     
            .or(Err("Invalid utf8 data"))?                                       
            .parse()                                                             
            .or(Err("Invalid decimal"))?;                                        
        let name_start = 11;                                                     
        let name_end = name_start + name_size;                                  
        let data_start = name_end;                                               
        let data_end = data_start + data_size;                                  
                                                                                 
        if data.len() < data_end {                                               
            return Err("Not enough data in response".to_string());              
        }                                                                        
                                                                                 
        let name = str::from_utf8(&data[name_start..name_end])                  
            .or(Err("Invalid ut8=f8 data"))?                                     
            .to_string();                                                       
                                                                                 
        let data = (&data[data_start..data_end]).to_vec();                       
                                                                                 
        Ok(File { name, data })                                                  
    }                                                                            
}                                                                               
```

plauche:
@davesims I'd prefer not to land this in master until we've tested against the actual hardware.

radio/nsl-duplex-d2/src/commands.rs
```diff
@@ -0,0 +1,42 @@
...
+}
+
+pub fn get_file_count() -> Command<FileCount> {
+    Command {
+        request: vec![0x47, 0x55, 0x47, 0x45, 0x54, 0x55, 0x46, 0x43],
+        parse: FileCount::parse,
+    }
+}
+
+pub fn get_state_of_health() -> Command<StateOfHealth> {
```
davesims: 💯 Love this, super clean.

creationix:
Ready for final review.

radio/nsl-duplex-d2/src/serial_comm.rs
```diff
@@ -0,0 +1,108 @@
...
+    let mut port = try!(serial::open("/dev/ttyUSB0"));
+    let settings: serial::PortSettings = serial::PortSettings {
+        baud_rate: serial::Baud38400,
+        char_size: serial::Bits8,
+        parity: serial::ParityNone,
+        stop_bits: serial::Stop1,
+        flow_control: serial::FlowNone,
+    };
+    try!(port.configure(&settings));
+
```
marshall: It looks like this coffee reconfigures / reopens a statically defined serial device + baudrate every time it writes data... Shouldn't serial device config be drive in device initialization?

radio/nsl-duplex-d2/src/serial_comm.rs
```diff
@@ -0,0 +1,108 @@
...
+    Ok(())
+}
+
+fn serial_receive() -> io::Result<Vec<u8>> {
+    use std::io::prelude::*;
+    use serial::prelude::*;
+
+    let mut ret_msg: Vec<u8> = Vec::new();
+    let mut port = try!(serial::open("/dev/ttyUSB0"));
```
marshall: same as above but for read

PR #30, opened 2018-02-12
Adding max values to target.config for AntS unit testing

CatGarab:
By default, the API for the ISIS antenna systems is configured with only two antennas and no secondary controller (to match the defaults of their options sheet).
The unit tests, however, need to have all four antennas enabled and both controllers in order to fully test the code (added in kubos/isis-api#17)

Since apparently yotta tests can't have their own `config.json` files, I'm putting the necessary configuration options into our test target, `x86-linux-native`.


CatGarab:
[Antenna](https://trello.com/c/SGugBeaT/191-antenna)

PR #36, opened 2018-02-20
GOMspace NanoPower API

CatGarab:
**PR Overview**:

- Adding an API for communication with the GOMspace NanoPower P31u

**Documentation**:

Inline documentation is done.
[User docs will be added in a future PR...](https://trello.com/c/bFzkuhdr/245-nanopower-api-user-guide)

**Testing**:

#44


eps/nanopower-api/nanopower-api/nanopower-api.h
```diff
@@ -0,0 +1,336 @@
...
+/**
+ *  @name GOMspace NanoPower config.json configuration options and default
+ * 		  values
+ */
+/**@{*/
+/**
+ * I2C bus the EPS is connected to
+ */
+#ifdef YOTTA_CFG_EPS_NANOPOWER_I2C_BUS
+#define EPS_I2C_BUS YOTTA_CFG_EPS_NANOPOWER_I2C_BUS
```
plauche: Some early feedback - I would highly suggest we start moving away from these compile time defines and start using run time configurable options. These compile time options work great when everything is built with yotta and make a ton of sense in the monolithic RTOS world, but much less so when combining Rust+C across many small individually purposed binaries.

CatGarab: :+1: Makes sense to me. There will need to be a larger discussion to figure out how it should be implemented

CatGarab:
[Nanopower P31U](https://trello.com/c/3JtPCCSh/192-nanopower-p31u)

eps/nanopower-api/source/nanopower.c
```diff
@@ -0,0 +1,737 @@
...
+
+    packet.cmd = SET_CONFIG1;
+
+    packet.sys_config.ppt_mode = config->ppt_mode;
+    packet.sys_config.battheater_mode = config->battheater_mode;
+    packet.sys_config.battheater_low = config->battheater_low;
+    packet.sys_config.battheater_high = config->battheater_high;
+    for (int i = 0; i < 8; i++)
+    {
+        packet.sys_config.output_normal_value[i] = config->output_normal_value[i];
```
plauche: The eps doesn't require all these parameters to be big endian?

eps/nanopower-api/source/nanopower.c
```diff
@@ -0,0 +1,737 @@
...
+    buff->curin[2] = be16toh(body->curin[2]);
+    buff->cursun = be16toh(body->cursun);
+    buff->cursys = be16toh(body->cursys);
+    for (int i = 0; i < 6; i++)
+    {
+        buff->curout[i] = be16toh(body->curout[i]);
+    }
+    for (int i = 0; i < 8; i++)
+    {
+        buff->output[i] = body->output[i];
```
plauche: Why no `be16toh` here?

eps/nanopower-api/source/nanopower.c
CatGarab: Most of them are only a single byte, so don't need to be changed.

eps/nanopower-api/source/nanopower.c
CatGarab: Each `output` element is a single byte, so endianness doesn't matter

eps/nanopower-api/source/nanopower.c
```diff
@@ -0,0 +1,737 @@
...
+        perror("Failed to rejoin EPS watchdog thread");
+        return EPS_ERROR;
+    }
+
+    handle_watchdog = 0;
+
+    return EPS_OK;
+}
+
+KEPSStatus k_eps_passthrough(const uint8_t * tx, int tx_len, uint8_t * rx,
```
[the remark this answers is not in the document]
CatGarab: `passthrough` just didn't have any error checking, for reasons which I don't remember. I changed it to just be a call through to `transfer`, which matches the other APIs

creationix:
I was going to comment on endianess when I saw the packed structs being used for network packets, but then I saw it was already discussed and best I can all, all the multi-byte values are handling the endianess concerns.

PR #40, opened 2018-03-05
Trimming Dockerfile

plauche:
Removing legacy BBB target and docker dependencies.


plauche:
:rotating_light: Once this PR is merged into master it will break the Circle check for all other branches until they update with master :rotating_light:

PR #43, opened 2018-03-06
Service outline doc

jacoffey3:
Created the actual GraphQL Schema for the services. This will be subject to change pending implementation.


jacoffey3:
[Service Outline Docs](https://trello.com/c/as6OVYis/213-service-outline-docs)

docs/dev-docs/service-outline-guide.rst
```diff
@@ -0,0 +1,375 @@
...
+
+General Hardware Service
+-------------------------
+
+A general hardware service is a service for any piece of hardware that does not fit into any of the other categories. These queries/mutations will be expected to be present regardless of the hardware. All other service outlines build on top of what is present here. 
+
+GraphQL Schema::
+
+    type Query {
+        ack(): String
```
plauche: Should `ack` be a `Boolean` instead of `String`?

docs/dev-docs/service-outline-guide.rst
```diff
@@ -0,0 +1,375 @@
...
+        power(): PowerState
+        config(): String
```
plauche: Should `config` be a `String` or a `Config` type specific to the hardware?

docs/dev-docs/service-outline-guide.rst
```diff
@@ -0,0 +1,375 @@
...
+    }
+    
+    type ConfigureHardwarePayload implements MutationResult {
+        errors: [String]
+        success: Boolean
+        config: String
+    }
+    
+    input ConfigureHardwareInput {
+        config: String
```
plauche: Should this be just a string or a type which is full of hardware specific options?

docs/dev-docs/service-outline-guide.rst
```diff
@@ -0,0 +1,375 @@
...
+
+ADCS Service
+------------
+
+The ADCS service outline and all following service outlines aim to abstract just the telemetry items and commands that are useful for mission logic. If you need a certain telemetry item for your mission application, please let us know! 
+
+Additional GraphQL Schema::
+
+    type Query {
+        mode(): String
```
plauche: Mode might be an `Enum` specific to the ADCS used?

docs/dev-docs/service-outline-guide.rst
jacoffey3: TBD. We can't know this until we know how it's going to be implemented.

docs/dev-docs/service-outline-guide.rst
jacoffey3: No, Ack is the actual last command received.

docs/dev-docs/service-outline-guide.rst
jacoffey3: This will probably change later, after we figure out what the actual implementation is.

jacoffey3: The idea is the Config will be "nominal" or some special tag, so it's more like an ENUM, but we don't know what it should be yet.

PR #44, opened 2018-03-06
Testing for NanoPower API

CatGarab:
Adding unit tests and an integration test for the NanoPower API

### DIFF
--- a/eps/nanopower-api/test/nanopower/sysfs.c
+++ b/eps/nanopower-api/test/nanopower/sysfs.c
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2017 Kubos Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


CatGarab:
[Nanopower P31U](https://trello.com/c/3JtPCCSh/192-nanopower-p31u)

eps/nanopower-api/test/nanopower/sysfs.c
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/* Returns number of bytes "written" or -1 on failure */
+ssize_t __wrap_write(int fd, const char * buf, size_t count)
+{
+    /* Verify that we're sending the correct command */
+    uint8_t cmd = buf[0];
+    check_expected(cmd);
+
+    if (count > 4)
```
plauche: Why only if `count > 4`?

CatGarab: It was something I set up specifically to check the messages which have endianness conversions. The smaller command packets (ex. `reset` and `ping`) don't have any multi-byte fields, so I just checked that their command was correct instead. It makes their test case code smaller and a little easier to read.

PR #46, opened 2018-03-07
Add most of nearspace commands (including `put_download_file`)

creationix:
This adds most the rest of the nearspace model commands (all except get geolocation
position estimate)

- [x] `get_state_of_health_for_modem() -> RadioResult<StateOfHealth>`
- [x] `get_uploaded_file_count() -> RadioResult<u32>`
- [x] `get_uploaded_message_count() -> RadioResult<u32>`
- [x] `get_download_file_count() -> RadioResult<u32>`
- [x] `get_uploaded_file() -> RadioResult<File>`
- [x] `get_uploaded_message() -> RadioResult<Message>`
- [x] `delete_download_files() -> RadioResult<u32>`
- [x] `delete_uploaded_files() -> RadioResult<u32>`
- [x] `delete_uploaded_messages() -> RadioResult<u32>`
- [x] `put_download_file(file: &File) -> RadioResult<bool>`
- [x] `get_alive() -> RadioResult<bool>`


creationix:
[Implement download file](https://trello.com/c/E1HL6FXW/233-implement-download-file)

creationix:
Ready for Review: @davesims @plauche

creationix:
I'm wondering if instead of parsing the sync bytes (`GU`) we should actually sync?  The difference would mean reading and throwing away data till we get to the `GU` sequence.

creationix:
We need to test on the actual radio.  It's unclear to me in the spec if we need to send a second sync after the put file command, but before the actual file.

radio/comms-service/src/main.rs
```diff
@@ -2,14 +2,25 @@ extern crate nsl_duplex_d2;
...
-    let file = radio.get_file().unwrap();
-    println!("File name: {}", file.name);
-    println!("File data: {:?}", file.body);
+    if count != 0 {
+        let file = radio.get_uploaded_file().unwrap();
+        println!("File name: {}", file.name);
+        println!("File data: {:?}", file.body);
+    }
+
+    println!("Sending file");
```
davesims: @creationix Should these stdout messages be sent to telemetry as more formal logs?

plauche: Sounds like we need a more formal logging -> telemetry story.

davesims: @plauche Yep. Chatted briefly with @pengwynn about this. MT has to decide the same thing, i.e., where to store strings as opposed to floats. stdout is 👍 for a placeholder now, let's revisit later.

PR #47, opened 2018-03-07
Rust testing

plauche:
- Added Rust dependencies to Docker
- Added script for setup/check/test rust
- Made `tools/kubos_link.py` faster
- Added scirpt to circle.yml
- Created Cargo workspace so we can test all Rust code from one place
- Fixed some broken dependencies


tools/rustbuild.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
CatGarab: Need execute permissions for this file

PR #53, opened 2018-03-20
Kubos Service for ISIS Antenna Systems

CatGarab:
Adding a Kubos service for ISIS antenna systems.

Misc:
- Some changes were made to the underlying API in order to support mocking for unit tests.
- With any luck, when the underlying API C code is converted to Rust, the service code should be unaffected


CatGarab:
[ISIS Antenna Systems Service](https://trello.com/c/CsQ7Alen/183-isis-antenna-systems-service)

services/isis-ants-service/src/model.rs
```diff
@@ -0,0 +1,336 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use failure::Fail;
+use isis_ants_api::{AntS, KANTSAnt, KANTSController, KI2CNum};
+use std::cell::RefCell;
+use std::io::Error;
```
plauche: Why are we using `std::io::Error` as our error type in `Result` returns instead of `AntsError`?

services/isis-ants-service/src/model.rs
```diff
@@ -0,0 +1,336 @@
...
+        subsystem
+    }
+
+    // Queries
+
+    pub fn get_arm_status(&self) -> Result<ArmStatus, Error> {
+        let (_errors, _success, deploy) = run!(self.ants.get_deploy(), self.errors);
+        let armed = deploy.unwrap_or_default().sys_armed;
+
+        let status = match armed {
```
plauche: nit: Could simplify to
```rust
       Ok(match armed {
            true => ArmStatus::Armed,
            false => ArmStatus::Disarmed,
        })
```

services/isis-ants-service/src/model.rs
```diff
@@ -0,0 +1,336 @@
...
+                    power: state,
+                    errors,
+                    success,
+                })
+
+            } 
+            _ => {
+                push_err!(self.errors, "controlPower: Invalid power state".to_owned());
+
+                Ok(ControlPowerResponse {
```
plauche: I'm sure this is related to the need for the `run!` macro, but is there a reason why we are storing a `true/false` status inside of the `Ok`, instead of using `Ok` for success and `Err` for failure?

services/isis-ants-service/S20isis-ants-service
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+NAME=isis-ants-service
```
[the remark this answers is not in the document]
CatGarab: Ultimately probably in the `kubos-linux-build` repo. However, the corresponding package hasn't been created yet (and will now be tabled), so I wanted to make sure it existed (and worked) in some form.

services/isis-ants-service/src/model.rs
CatGarab: In general, it's because the service doc specifies that there should be `errors` and `success` fields available for each of the mutations. This particular case could probably just be an error

services/isis-ants-service/src/macros.rs
```diff
@@ -0,0 +1,307 @@
...
+///     assert_eq!(errors, "TopError: top, RootError: root");
+///     assert_eq!(success, false);
+///     assert_eq!(data, None);
+///     assert_eq!(
+///         vec!["test_func: TopError: top, RootError: root".to_owned()],
+///         master_err.borrow().clone()
+///     );
+/// }
+/// ```
+macro_rules! run {
```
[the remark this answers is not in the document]
CatGarab: Note to future self: Do this.

jacoffey3:
[ISIS Antenna service](https://trello.com/c/xuvZ4M6q/394-isis-antenna-service)

apis/isis-ants-api/src/ants.rs
```diff
@@ -33,10 +33,32 @@ pub enum AntsError {
 /// Custom result type for antenna operations
 pub type AntSResult<T> = Result<T, AntsError>;
 
+pub trait IAntS {
+    fn new(bus: &str, primary: u8, secondary: u8, ant_count: u8, timeout: u32) -> AntSResult<Self>
+    where
+        Self: ::std::marker::Sized;
```
plauche: Why was this `where Self: ::std::marker::Sized` necessary? Just curious.

CatGarab: If I remember correctly, Rust was totally okay with returning `Self` from the trait function, but if I wrapped it in a result, it suddenly threw "the trait `std::marker::Sized` is not implemented for `Self`". So the `where` statement fixes that.

PR #58, opened 2018-03-30
WIP: Initial implementation of udp-http-relay

creationix:
This service listens on a fixed UDP port and accepts HTTP requests as datagrams.  

It then acts as an HTTP reverse proxy and makes http requests to the local services that speak http.  The response is send back to the original UDP sender.

I tested this with the following setup:

```sh
 # Run this service
PORT=4000 luvi-tiny . --

# Run a socat proxy
socat tcp-listen:3001,fork,reuseaddr udp:localhost:4000

# Run an http server
hexes

# Make a curl query to hexes through socat and the relay
curl -i http://localhost:3001/ -H'Host: localhost:8080'
```


creationix:
Not needed anymore

PR #59, opened 2018-04-02
[WIP] GraphQL over UDP

plauche:
- Add UDP interface for GraphQL services
- Add kubos-service crate to abstract network interface and service creation

### DIFF
--- a/services/kubos-service/src/service.rs
+++ b/services/kubos-service/src/service.rs
@@ -1,0 +1,92 @@
+//
+// Copyright (C) 2017 Kubos Corporation
+//
+// Licensed under the Apache License, Version 2.0 (the "License")
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+use config::Config;
+use std::net::{SocketAddr, UdpSocket};
+use serde_json::to_string;
+use juniper::{execute, GraphQLType, RootNode, Variables};
+
+pub struct KubosService<'a, CtxFactory, Query, Mutation, CtxT>
+where
+    CtxFactory: Fn() -> CtxT + Send + Sync + 'static,
+    CtxT: 'static,
+    Query: GraphQLType<Context = CtxT> + Send + Sync + 'static,
+    Mutation: GraphQLType<Context = CtxT> + Send + Sync + 'static,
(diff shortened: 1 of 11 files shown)


plauche:
This is like 95% ready for review. There is a bit more doc work I want to do but reviewing should begin.

plauche:
@CatGarab @creationix @davesims do we have a strong preference on `json` vs `YAML` for our configuration file? I initially used `json` but looking back I think `YAML` is a bit better looking for human readable configs and has less editing overhead.

`json`:
```json
{
  "example-service": {
    "addr": {
      "ip": "127.0.0.1",
      "port": 8080
    },
    "i2c-bus": "/dev/ttyUSB0"
  }
}
```

`YAML`:
```yaml
- example-service: 
  - addr:
    - ip: 127.0.0.1
    - port: 8080
  - i2c-bus: /dev/ttyUSB0
```

plauche:
[UDP and Services Abstraction](https://trello.com/c/dLhCuUA8/285-udp-and-services-abstraction)

davesims:
@plauche Slight preference for YAML for readability but I could go either way. 🤷‍♂️

creationix:
JSON was never meant to be human written (supposedly).  Personally I'm a fan of toml for config files.  It looks like good 'ol INI, but has an actual spec and structure.

Surely rust has good toml parsers since that's what cargo uses for it's config.

creationix:
https://github.com/toml-lang/toml#comparison-with-other-formats

plauche:
I'm also good with toml. It has good Rust support http://alexcrichton.com/toml-rs/toml/index.html

services/kubos-service/src/service.rs
```diff
@@ -0,0 +1,188 @@
...
+            &self.context,
+        ) {
+            Ok((val, _errs)) => {
+                // Should do something with _errs
+                return to_string(&val).unwrap();
+            }
+            Err(_e) => {
+                "Error running query".to_string()
+                // Could also do this to retain the juniper error
+                // return serde_json::to_string(&e).unwrap(),
```
CatGarab: My preference is always to retain the underlying error

services/kubos-service/src/service.rs
```diff
@@ -0,0 +1,188 @@
...
+    fn process(&self, query: String) -> String {
+        match execute(
+            &query,
+            None,
+            &self.root_node,
+            &Variables::new(),
+            &self.context,
+        ) {
+            Ok((val, _errs)) => {
+                // Should do something with _errs
```
CatGarab: The errors here are things like "the field you requested doesn't exist", which is definitely important. (The Juniper examples that actually process the errors are [here](https://github.com/graphql-rust/juniper/blob/master/juniper/src/executor_tests/executor.rs), if you haven't seen them)

My suggestion would be to check if `errs` is empty; if not, then convert all of the errors to one big string and return that. (maybe instead of the `val` string? Seems weird to combine the two...)

PR #60, opened 2018-04-05
Trying to catch periodic CSP unit test failure

CatGarab:
Periodically the `test_client` test case in `libcsp/test/csp_if_socket_client.c` fails for no good reason. It's super irritating.

My best guess is that occasionally the CSP server isn't actually completely initialized by the time the client attempts to communicate with it, so I added an extra sleep before the client attempts to connect.

If I'm wrong, I also made changes to enable verbose debugging output for the native target so that all `csp_log_error` calls are output and will hopefully give us useful failure information.


CatGarab:
@marshall I tried that and it hangs the program. Under the covers `socket_init` ends up waiting for an initial client connection. The client is now waiting for the server's `socket_init` to complete, so it never starts a connection.

This PR is definitely more of a bandaid, but I rabbit-holed for too long trying to fix the root problem.

plauche:
Can we just remove the offending tests...since we aren't using that part of CSP anymore?

CatGarab:
@plauche That's definitely an option. My preference would be to leave the unit test until the code it tests is removed, just in case we return to it. If this PR doesn't work, then definitely delete the unit test, though.

PR #63, opened 2018-04-17
Kubos Service for MAI-400 ADACS

CatGarab:
Kubos service for use when interacting with an MAI-400 ADACS

A KLB package has been created in kubos/kubos-linux-build#90


CatGarab:
[MAI 400 Service](https://trello.com/c/5yUH0sp8/219-mai-400-service)

services/kubos-service/src/macros.rs
```diff
@@ -0,0 +1,299 @@
...
+///     let errors = process_errors!(chain);
+///     assert_eq!(errors, "TopError: top, RootError: root");
+///
+///     let errors = process_errors!(chain, '\n');
+///     assert_eq!(errors, "TopError: top\nRootError: root");
+/// }
+/// ```
+///
+#[macro_export]
+macro_rules! process_errors {
```
plauche: :100:

services/mai400-service/src/model.rs
```diff
@@ -0,0 +1,355 @@
...
+    pub errors: RefCell<Vec<String>>,
+    pub persistent: Arc<ReadData>,
+}
+
+impl Subsystem {
+    pub fn new(bus: String, data: Arc<ReadData>) -> Subsystem {
+        println!("New Subsystem");
+
+        let connection = Connection::new(bus.clone());
+        let mai = MAI400::new(connection);
```
plauche: Looks like we have two different instances of `MAI400` each with their own `Connection`. How is it handled when both of these are attempting to communicate with the hardware at the same time?

services/mai400-service/src/model.rs
```diff
@@ -0,0 +1,427 @@
...
+impl Subsystem {
+    pub fn new(bus: &'static str, data: Arc<ReadData>) -> Subsystem {
+        let connection = Connection::new(bus);
+        let mai = MAI400::new(connection);
+
+        let data_ref = data.clone();
+
+        let (sender, receiver) = channel();
+
+        spawn(move || read_thread(MAI400::new(Connection::new(bus)), data_ref, sender));
```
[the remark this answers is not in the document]
CatGarab: For the MAI400, the main thread only writes and the read thread only reads. There will never be an instance where the main thread needs to write and then immediately read because none of the commands have responses. The only way to check if they actually succeeded is to look at the `cmd_valid_cntr` and `cmd_invalid_cntr` values in the next standard telemetry packet (which is output by the MAI every 1/4 sec) and see which one incremented.

I coded it this way because Stream can't be shared across threads (I think there's an Rc or something under the covers that prevents it). In retrospect, I could probably wrap it in a mutex and then we could guarantee to only be doing one I/O operation at a time. It shouldn't make a difference for this PR, but might be better/safer for other services going forward.

PR #64, opened 2018-04-23
MAI400 Integration Test

CatGarab:
Integration test for the MAI400 using the API created in #61


CatGarab:
[MAI 400 API](https://trello.com/c/jGukXX8n/218-mai-400-api)

PR #65, opened 2018-04-24
Generalized Rust UART HAL

greer184:
This PR provides a more generalized HAL interface to read and write across serial ports. From the friendly folks at AFIT.


hal/rust-hal/rust-uart/src/lib.rs
```diff
@@ -0,0 +1,81 @@
+//! A generalized HAL for communitcating over serial ports
```
plauche: nit: communicating spelling

hal/rust-hal/rust-uart/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+pub struct Connection {
+    pub stream: Box<Stream>,
+}
+
+/// Errors that occur while reading from and writing to stream
+pub type UartResult<T> = Result<T, std::io::Error>;
+
+impl Connection {
+    /// Constructor to create connection from path
+    pub fn new(bus: String) -> Connection {
```
plauche: It would be easier for mocking purposes if the `Connection` constructor took a `Box<Stream>` parameter, rather than always creating a `SerialStream` internally. You could add a convenience function which takes a `bus: String` parameter and returns a `Connection` made with a `SerialStream`, that way it is easy to create the "actual" connections, but it isn't hard coded inside of the constructor.

greer184:
Yeah, that makes sense so you can create SerialStreams externally. I went ahead and switched the old constructor into a convenience method.

Cargo.toml
```diff
@@ -13,6 +13,7 @@ members = [
 "services/kubos-service",
 "services/iobc-supervisor-service",
 "test/integration/linux/isis-ants",
+"hal/rust-hal/rust-uart",
```
plauche: Can you put this in alphabetical order?

hal/rust-hal/rust-uart/src/lib.rs
```diff
@@ -0,0 +1,86 @@
+//! A generalized HAL for communicating over serial ports
```
plauche: Can you add `#![deny(missing_docs)]`?

hal/rust-hal/rust-uart/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+extern crate serial;
+
+use std::io::prelude::*;
+use serial::prelude::*;
+
+/// Wrapper for UART stream
+pub struct Connection {
+    pub stream: Box<Stream>,
```
plauche: Can you document this public member?

plauche:
Hey @greer184! It looks like I didn't have the correct CircleCI settings for building PRs from forks. Would you mind closing this PR and opening a new identical one? I can't find a way to create a build for this existing PR but I think Circle will catch the new one.

PR #66, opened 2018-04-24
Example Client for Kubos Services

CatGarab:
Kubos service queries can get quite long and I got tired of having to type out full shell commands, so I created a UDP client program to use instead.

Modeled after the new `kubos-service` crate, the example takes in a config file which specifies the host IP, the service IP, and the name of the file containing the query to send.


examples/udp-service-client/config.toml
```diff
@@ -0,0 +1,9 @@
+query_file = "query.txt"
+
+[host]
+ip = "10.0.2.15"
```
plauche: nit: It might be better if the example `config.toml` used `127.0.0.1` as the host ip. This would make it so that users would only need to change the service ip to get going.

examples/udp-service-client/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+fn configure_client() -> (SocketAddr, SocketAddr, String) {
+    let args: Vec<String> = env::args().collect();
+
+    let config: Config = if args.len() == 2 {
+        let mut raw = String::new();
+        let filename = &args[1];
+        match File::open(filename)
+            .map(|mut f| f.read_to_string(&mut raw))
+            .map(|_| toml::from_str(&raw)) {
+            Ok(toml) => toml.unwrap(),
```
plauche: nit: Run `rustfmt` over this

examples/udp-service-client/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    // Allocate the buffer and read the response
+    let mut buf = vec![0u8; len].into_boxed_slice();
+    let result = socket.recv_from(&mut buf);
+    match result {
+        Ok((amt, src)) => {
+            println!("Received {} bytes from {}", amt, src);
+            let mut v: serde_json::Value = serde_json::from_slice(&buf[0..(amt)]).unwrap();
+            // If there's  'msg' field in the returned JSON, then our query went (atleast mostly) successfully.
```
plauche: Currently if someone builds a service using the `kubos-service` crate every query will have a `msg` field, but it will be empty if the query fails - [relevant source](https://github.com/kubos/kubos/blob/88195073d0d13c9a29517ef98750e4b7707c0150/services/kubos-service/src/service.rs#L177).
The result is that the client will panic and crash :cry:

PR #70, opened 2018-04-27
Removing occasionally faulty unit test

CatGarab:
The `csp_if_socket_client` unit test occasionally fails when run by CircleCI. At a guess, the client attempts to connect to the server before the server has finished initializing. Previous attempts have been made to fix it, but the problem still occurs.

The code being tested by this file isn't currently used by any other Kubos code, so we've deemed it safe for removal. Moving forward, we'll be removing the entire `libcsp` folder.


CatGarab:
[Pull out defunct unit tests](https://trello.com/c/V8l0uCMr/134-pull-out-defunct-unit-tests)

PR #79, opened 2018-05-10
Force downgrade pip to <v10

CatGarab:
There's a dependency incompatibility issue which causes the `RUN pip install -r https://raw.githubusercontent.com/kubos/kubos-cli/master/requirements.txt` command to fail: https://github.com/ARMmbed/yotta/issues/835
Forcing the pip install to remain at v9 until the issue is resolved


CatGarab:
[Fix Docker/Vagrant provisioning pip version issue](https://trello.com/c/QU2xTq8c/312-fix-docker-vagrant-provisioning-pip-version-issue)

PR #81, opened 2018-05-14
Rust UART HAL Updates

CatGarab:
This PR adds a few features which I found that I needed while working on the NovAtel OEM6 API in kubos#74:

- Updated `Connection::from_path` to have two new parameters: settings and timeout.
- [Failure](https://github.com/rust-lang-nursery/failure) is our error-handling crate-of-choice, so converted the current error type into a Failure error type
- Changed the code to open the connection to the serial port once, during the `SerialStream::new()` call, rather than during each `write()` and `read()` call, so that messages which require multiple read calls (ex. where we need to read a header to then get the length of the whole message) will work
- Tweaked the `read` calls to add a timeout parameter. There's currently no way to do an indefinite `read` call with the [serial crate](https://github.com/dcuddeback/serial-rs), so my intent with this is to set a longer timeout (ex. one second) for reads which need to wait for data to come in, but then make the timeout short (ex. 50 milliseconds) for reads which should return almost instantly. This is especially useful for multi-part reads, where I want to wait for a message, read only the header, then read the message body.
- Marked the `Stream` trait as `Send` compatible, so that we can use a stream in multi-threaded code (note: doing this properly requires the use of a mutex, but that burden will be placed on the caller since this library isn't inherently multi-threaded)
- Created `MockStream` so that we can create unit tests. We were using [Double](https://github.com/DonaldWhyte/double), but that isn't multi-thread compatible (they use `Rc`s :disappointed:)

### DIFF
--- a/hal/rust-hal/rust-uart/src/lib.rs
+++ b/hal/rust-hal/rust-uart/src/lib.rs
@@ -21,66 +21,22 @@
 extern crate failure;
 extern crate serial;
 
+mod error;
 pub mod mock;
 #[cfg(test)]
 mod tests;
 
+use error::*;
 use std::io::prelude::*;
 use serial::prelude::*;
 use std::time::Duration;
 use std::cell::RefCell;
-use std::error::Error;
 
 /// Wrapper for UART stream
 pub struct Connection {
     /// Any boxed stream that allows for communication over serial ports
     pub stream: Box<Stream>,
 }
-
-/// Custom errors for UART actions
-#[derive(Fail, Display, Debug, Clone, PartialEq)]
-pub enum UartError {
-    /// Catch-all error case
-    #[display(fmt = "Generic Error")]
-    GenericError,
-    #[display(fmt = "Serial Error: {}", description)]
-    /// An error was thrown by the serial driver
-    SerialError {
-        /// The underlying error type
-        cause: serial::ErrorKind,
-        /// Error description
-        description: String,
-    },
-    #[display(fmt = "IO Error: {}", description)]
-    /// An I/O error was thrown by the kernel
-    IoError {
-        /// The underlying error type
-        cause: std::io::ErrorKind,
-        /// Error description
-        description: String,
-    },
-}
-
-impl From<std::io::Error> for UartError {
-    fn from(error: std::io::Error) -> Self {
-        UartError::IoError {
-            cause: error.kind(),
-            description: error.description().to_owned(),
-        }
-    }
-}
-
-impl From<serial::Error> for UartError {
-    fn from(error: serial::Error) -> Self {
-        UartError::SerialError {
-            cause: error.kind(),
-            description: error.description().to_owned(),
-        }
-    }
-}
-
-/// Errors that occur while reading from and writing to stream
-pub type UartResult<T> = Result<T, UartError>;
 
 impl Connection {
     /// Constructor to creation connection with provided stream
--- a/hal/rust-hal/rust-uart/src/lib.rs
+++ b/hal/rust-hal/rust-uart/src/lib.rs
@@ -26,7 +26,7 @@ pub mod mock;
 #[cfg(test)]
 mod tests;
 
-use error::*;
+pub use error::*;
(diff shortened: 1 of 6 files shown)


CatGarab:
[Novatel OEM GPS](https://trello.com/c/LU7jpmgz/283-novatel-oem-gps)

srjustice:
@CatGarab First off, we 100% agree with all the changes! We realized the shortcomings of the original HAL that we gave you guys pretty quickly after it was merged into master, and we actually made a lot of the same changes to our local copy that you made in this PR, haha. Glad we are on the same page! 

I did have a couple of questions though. First,  I noticed that you use `unwrap()` on the result of `try_borrow_mut()` in both the `write` and `read` functions for `SerialStream`. Won't that cause a panic, which will kill whatever thread this code is running in?

 Also, just out of curiosity, what made you wanna go with the anonymous structs for the `IoError` and `SerialError` variants of `UartError`. Couldn't you replace those with `SerialError(serial::Error)` and `IoError(std::io::Error)` and just pass the actual errors themselves up the line? Then, for example, your `From` implementation for `serial::Error` would basically just be:
```
impl From<serial::Error> for UartError {
    fn from(error: serial::Error) -> Self {
        UartError::SerialError(error)
    }
}
```
It seems like that would be a little cleaner and save you from the overhead of having to clone the descriptions.

We love the `MockStream`! That will be very useful! Too bad `Double` isn't multi-thread compatible! Also, I am very happy to move to only opening the serial port once!

hal/rust-hal/rust-uart/src/mock.rs
```diff
@@ -0,0 +1,240 @@
+//
```
plauche: This is :100:

hal/rust-hal/rust-uart/src/mock.rs
```diff
@@ -0,0 +1,240 @@
...
+        } else {
+            self.write.result.clone()
+        }
+    }
+
+    fn read(&self, len: usize, _timeout: Duration) -> UartResult<Vec<u8>> {
+        if let Some(ref output) = self.read.output {
+            let mut response: Vec<u8> = vec![0; len];
+
+            match output.borrow_mut().read_exact(response.as_mut_slice()) {
```
plauche: Do we want `try_borrow_mut` or `borrow_mut` here?

hal/rust-hal/rust-uart/src/lib.rs
```diff
@@ -1,40 +1,58 @@
...
 #![deny(missing_docs)]
 
 //! A generalized HAL for communicating over serial ports
+#[macro_use]
+extern crate failure;
 extern crate serial;
 
-use std::io::prelude::*;
+mod error;
+pub mod mock;
```
plauche: Should we also add the `#[cfg(test)]` attribute to `pub mod mock`?

CatGarab: That actually only applies to the crate you're actively building/testing. If you try to use it in a different crate and run that one's tests, it won't find the `mock` module, presumably because it's not included.

If you can find a way to actually do it, I'm all ears :)

hal/rust-hal/rust-uart/src/lib.rs
```diff
@@ -43,42 +61,54 @@ impl Connection {
...
+    }
 }
 
 // Read and write implementations for the serial stream
 impl Stream for SerialStream {
     fn write(&self, data: &[u8]) -> UartResult<()> {
-        let mut port = serial::open(self.bus.as_str())?;
-
-        port.configure(&self.settings)?;
+        let mut port = self.port.try_borrow_mut().unwrap();
```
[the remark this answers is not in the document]
CatGarab: I need a linter to catch these for me >.>

hal/rust-hal/rust-uart/src/mock.rs
CatGarab: I don't think that there should ever be simultaneous read calls in unit tests, so if this ever fails, I definitely want it to panic.

PR #82, opened 2018-05-16
[WIP] Pumpkin mcu service

jacoffey3:


jacoffey3:
[Pumpkin MCU service](https://trello.com/c/mVlkY2ub/314-pumpkin-mcu-service)

services/pumpkin_mcu_service/pumpkin_mcu_api/test_mcu_api.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+
+# Copyright 2017 Kubos Corporation
```
marshall: nit: it is 2018

services/pumpkin_mcu_service/pumpkin_mcu_api/test_mcu_api.py
```diff
@@ -0,0 +1,33 @@
...
+
+import mcu_api
+
+MODULES = mcu_api.CONFIG_DATA['modules']
+
+for module in MODULES:
+    module = str(module)
+    address = MODULES[module]['address']
+    if address == 0:
+        print('\nModule not configured: ' + module + '\n')
```
marshall: why do you prepend newlines here and on line 21, and append a newline on 22? why not always just append? is it because you aren't adding newlines on line 33?

services/pumpkin_mcu_service/pumpkin_mcu_api/test_mcu_api.py
```diff
@@ -0,0 +1,33 @@
...
+Testing mcu_api
+"""
+
+import mcu_api
+
+MODULES = mcu_api.CONFIG_DATA['modules']
+
+for module in MODULES:
+    module = str(module)
+    address = MODULES[module]['address']
```
marshall: how does this work when `module` isn't already a `str`? maybe a better question to ask, why do you need to convert `module` to a `str`?

services/pumpkin_mcu_service/service.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+
+# Copyright 2017 Kubos Corporation
```
marshall: nit: it is 2018

services/pumpkin_mcu_service/service.py
```diff
@@ -0,0 +1,22 @@
...
+# See LICENSE file for details.
+
+"""
+Boilerplate main for service application.
+"""
+
+from service import app
+import json
+
+# I know this is hacky...I need help with how to manage config files
```
marshall: I think the best way to solve this problem is to just put the configuration in code, it won't be changeable for users anyway right?

services/pumpkin_mcu_service/service/schema.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python
+
+# Copyright 2017 Kubos Corporation
```
marshall: nit: party like it's 2018

services/pumpkin_mcu_service/service/schema.py
```diff
@@ -0,0 +1,104 @@
...
+    
+    def resolve_rawRead(self, info, module, count):
+        """
+        Reads number of bytes from the specified MCU
+        """
+        if module not in MODULES:
+            raise KeyError('Module not configured',module)
+        address = MODULES[module]['address']
+        if address == 0:
+            raise ValueError('Module not present',module)
```
marshall: great code here 👍

services/pumpkin_mcu_service/service/schema.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Mutation(graphene.ObjectType):
+    """
+    Creates mutation endpoints exposed by graphene.
+    """
+
+    passthrough = Passthrough.Field()
+
+schema = graphene.Schema(query=Query, mutation=Mutation)
```
marshall: nit: you use spaces around your `kwargs` on line 92, but not here. I don't really care which you use but at least be consistent :)

services/pumpkin_mcu_service/service.py
marshall: also just realizing, you have two different modules reading and accessing this data, which is a big no-no. IMO the reading of this data (whether code or JSON based) should be isolated in code to one place, not duplicated

services/pumpkin_mcu_service/service/schema.py
```diff
@@ -0,0 +1,104 @@
...
+from models import *
+
+from pumpkin_mcu_api import mcu_api
+
+# Get what modules are present from config file
+MODULES = mcu_api.CONFIG_DATA['modules']
+
+# Local subsystem instance for tracking state
+# May not be neccesary when tied into actual hardware
+#_local = Subsystem(power_on=False)
```
marshall: nit: dead code should be removed altogether not just commented

services/pumpkin_mcu_service/service.py
jacoffey3: They are 2 different config files, I don't access the same file. 
And I could, but it seemed like it would bloat the code. Also then becomes confusing for users as what is meant  to be changed and what isn't.

services/pumpkin_mcu_service/service/schema.py
jacoffey3: 😆

services/pumpkin_mcu_service/service/schema.py
jacoffey3: I agree, when it's finished :)

services/pumpkin_mcu_service/pumpkin_mcu_api/test_mcu_api.py
jacoffey3: This is a integration test file for my personal use...will not be the same when finished. I wanted it to be easy to read.

jacoffey3: @jacoffey3

PR #83, opened 2018-05-17
MAI400 API: Adding mutex usage for thread safety

CatGarab:
Opening a connection to an underlying UART stream is bad. Shame on me.

The connection will now be wrapped in a mutex so that it will be impossible for more than one I/O operation to take place at a time.

`Double` doesn't support `Send`, so converting the unit tests to use our new MockStream object.

Misc: Also adding random things needed by the MAI400 service in #63


CatGarab:
[MAI 400 Service](https://trello.com/c/5yUH0sp8/219-mai-400-service)

apis/mai400-api/src/mai400.rs
```diff
@@ -310,14 +336,58 @@ impl MAI400 {
...
-    > {
-        let mut msg = self.conn.read()?;
+    ) -> MAIResult<(
+        Option<StandardTelemetry>,
+        Option<RawIMU>,
+        Option<IREHSTelemetry>,
+    )> {
+        let mut msg = vec![];
+        loop {
+            {
```
plauche: Is it possible for the mai to get in a state where it's not sending data and leave us stuck in this loop?

plauche: :+1:

PR #85, opened 2018-05-21
OEM6 Service: Phase 1

CatGarab:
Adding the `issueRawCommand` and `configureHardware` mutations.

Other changes:
- Tweaking API's `request_errors` function to add `hold` arg, allowing logging to persist through `unlog_all` requests
- Changing OEMError::UartError display text so it isn't just printing the underlying error twice
- Changing the `service_new!` mocking setup macro to use actual `read_thread` function in order to go through full send-command->get-response logic


CatGarab:

services/novatel-oem6-service/src/tests/schema/test_data.rs
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright (C) 2018 Kubos Corporation
```
plauche: :100: moving the test data in here makes the tests much more readable. You can probably move most or all of the data you set in `set_input` and `set_output` to here.

PR #88, opened 2018-05-23
Python Service Package

plauche:
- [x] Create package for creating python services
  - [x] Same configuration interface as kubos-service crate
  - [x] UDP interface
  - [x] HTTP interface


libs/kubos_python_service/config.py
```diff
@@ -0,0 +1,43 @@
...
+    raw = []
+    
+    def __init__(self, name):
+        args = get_args(name)
+        if args.config:
+            path = args.config
+        else:
+            path = DEFAULT_PATH
+        try:
+            data = toml.load(path)
```
jacoffey3: It is necessary to put this info in a config file? It's little enough information that I'd rather just have the DEFAULT variables be what it uses.

plauche: This [pattern](https://github.com/kubos/kubos/blob/master/services/kubos-service/src/config.rs) was established in the `kubos-service` crate used to create rust services. We will have a master config file that holds the ip, port and additional configuration data for all running services. It is much simpler to find and change configuration data for multiple interacting services when it is stored in one central location, rather than hard coded in each service.

The intent is that this python package could be used by any python service, so following the same pattern allows it to be managed in the same way as the rust services.

plauche:
[Python UDP Interface](https://trello.com/c/Urx9Gerw/310-python-udp-interface)

jacoffey3:
Should we be including the http portion?

plauche:
This PR is deprecated in favor of #97 .

PR #90, opened 2018-05-29
OEM6 Service: Phase 3

CatGarab:
Adding position queries:

- `lockStatus` - The current validity of the position data
- `lockInfo` - The actual current position data

Additional work:
- In order to get the time information, I had to update the API to grab the timestamp from the header and then add it to the `BestXYZLog` structure

TODOs:
- There's one `TODO` present that I could use feedback on


CatGarab:

services/novatel-oem6-service/src/main.rs
```diff
@@ -66,16 +66,17 @@ mod schema;
 mod tests;
 
 use kubos_service::{Config, Service};
-use model::Subsystem;
+use model::{LockData, Subsystem};
 use novatel_oem6_api::OEMResult;
 use schema::{MutationRoot, QueryRoot};
+use std::sync::Arc;
 
 // TODO: CHANGE THE UART BUS TO UART4!!!
```
plauche: Re this todo - we could also pull the uart bus from the config file.

PR #92, opened 2018-05-30
OEM6 Service: Phase 4

CatGarab:
Adding queries:
- `power`
- `systemStatus`

Additional work:
- Converted the message header `recv_status` field from a raw u32 into bit fields
- Added system time, time status, and receiver status to all log messages in the API. They're not all used for each log within the service, but they're useful fields to have in general
- There a several functions which get and use the VersionInfo log. Changed the code to make one central `get_version_log()` function

### DIFF
--- a/services/novatel-oem6-service/src/objects.rs
+++ b/services/novatel-oem6-service/src/objects.rs
@@ -339,6 +339,20 @@ graphql_object!(LockInfo: () | &self | {
     }
 });
 
+/// Response field for 'power' query
+#[derive(GraphQLEnum, Clone, Eq, PartialEq, Debug)]
+pub enum PowerState {
+    On,
+    Off,
+}
+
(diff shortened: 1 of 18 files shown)


CatGarab:

services/novatel-oem6-service/src/objects.rs
```diff
@@ -339,6 +339,35 @@ graphql_object!(LockInfo: () | &self | {
...
+/// Response field for 'power' query
+#[derive(GraphQLEnum, Clone, Eq, PartialEq, Debug)]
+pub enum PowerState {
+    On,
+    Off,
+}
+
+/// Response fields for 'power' query
+#[derive(GraphQLObject)]
+pub struct GetPowerResponse {
```
plauche: As this file steadily approaches 500 lines we may want to break it up into several smaller ones.

PR #97, opened 2018-06-06
Python package for building services.

plauche:
Provides abstraction for creating a UDP or HTTP/TCP based service in Python. Aims to provide the same functionality as the Rust `kubos-service` crate.


plauche:
[Python UDP Interface](https://trello.com/c/Urx9Gerw/310-python-udp-interface)

jacoffey3:
It still requires an argument to start. Can you make it default to the default config file location if there aren't any args?

plauche:
@jacoffey3 that optional config arg should be fixed now

libs/kubos-service/requirements.txt
```diff
@@ -0,0 +1,4 @@
+graphene==2.0.1
+Flask==0.12.2
+Flask-GraphQL==1.4.1
```
CatGarab: Flask is only a requirement if you're using the HTTP version. I think this will cause problems when attempting to install `kubos-service` into the current KubOS setup because we no longer have Flask installed. (Note: I have not actually tested this to verify)

Is there a way to make this a conditional requirement or something?

examples/python-service/service.py
```diff
@@ -8,16 +8,19 @@
...
 
-app = app.create_app()
-app.run(host=cfg['APP_IP'], port=cfg['APP_PORT'])
+from kubos_service import udp_service
+
+# Start a udp service with optional context
+# udp_service.start(config, schema, {'bus': '/dev/ttyS3'})
+
+# Start a udp service
+udp_service.start(config, schema)
```
CatGarab: I think this might be a little more readable if it was just broken into two files. udp_service.py and http_service.py

libs/kubos-service/requirements.txt
jacoffey3: I agree with you here. I don't think it's worth doing a conditional requirement (mainly because I don't think that's possible), but breaking it into 2 services would be easy.

examples/python-service/service.py
jacoffey3: I disagree. As someone using this I would rather have it in the same example file because the actual use case is I'm making a service, and I want to use graphiql while I'm actively developing and debugging, and then just switch to the UDP for final integration on the stack.

CatGarab: :+1: Okay, I'll defer to actual user experience :)

libs/kubos-service/requirements.txt
plauche: Looks like it isn't unusual to have an `optional-requirements.txt` that the user has to explicitly install from.

libs/kubos-service/optional-requirements.txt
```diff
@@ -0,0 +1,7 @@
+## Optional Dependencies for Additional HTTP Functionality
+##
+## To install, run:
+##     $ pip install -r optional-requirements.txt
```
CatGarab: :100:

PR #100, opened 2018-06-11
Adding Rust doc generation to gendocs.py

CatGarab:
Our new Rust crates have lots of documentation which should be included in our official docs. This PR starts to integrate them by using Rust's doc gen (via `cargo kubos -c doc` to handle our custom C/Rust combo crates) and then copying the resulting HTML files into our usual `html` output folder.

Our existing docs are then able to reference them using relative links. An example has been included in this PR.

For an example doc link: http://docs.kubos.co/beta/services/core-services.html
To look at all the Rust docs: http://docs.kubos.co/beta/rust-docs/telemetry_service/index.html

Notes: 
- I looked at [Doxidize](https://github.com/steveklabnik/doxidize), but was unable to get it to quickly work. It's a very new tool, so might still be good to return to at some point.
- There isn't currently a way to generate a master index.html file for all of the Rust crates, so we can't just have one easy pointer to the Rust docs. [There's an open feature request for it, though](https://github.com/rust-lang/rust/issues/16103)

### DIFF
--- a/docs/services/core-services.rst
+++ b/docs/services/core-services.rst
@@ -5,7 +5,7 @@ The core services are all the services that provide critical flight software cap
 
 Upcoming services are:
 
- - Telmetry Database Service
+ - `Telmetry Database Service <../rust-docs/telemetry_service/index.html>`_
  - Communication Service
  - Shell Service
  - File Service
--- a/docs/services/core-services.rst
+++ b/docs/services/core-services.rst
@@ -5,7 +5,11 @@ The core services are all the services that provide critical flight software cap
 
 Upcoming services are:
 
- - `Telmetry Database Service <../rust-docs/telemetry_service/index.html>`_
+ - |telem-db|
  - Communication Service
  - Shell Service
  - File Service
+ 
+ .. |telem-db| raw:: html
+ 
+    <a href="../rust-docs/telemetry_service/index.html" target="_blank">Telemetry Database Service</a>
--- a/tools/gendocs.py
+++ b/tools/gendocs.py
@@ -3,6 +3,7 @@ import argparse
 import subprocess
 import os
 import shutil
+from distutils import dir_util
 
 GENERATE_XML = """
 (cat {0};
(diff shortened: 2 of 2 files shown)


CatGarab:
[Integrate Rust docs into main docs](https://trello.com/c/EhSNE9dZ/243-integrate-rust-docs-into-main-docs)

plauche:
We should investigate a way to maintain a view/link back to the main kubos docs when navigating the rust docs. Right now the only way is using the browser's back button (assuming they didn't open in a new tab/window).

Two ideas:
- Clicking on a link to a rust doc will open an iframe in that right panel containing the rust doc
- Injecting a link to the top level kubos docs into each of the rust doc html pages

jacoffey3:
@plauche how about we have it where it just opens in a new tab?

CatGarab:
Ok @plauche @jacoffey3 I updated it to open the links in a new tab/window (exact behavior is browser-dependent)

I tried using an iframe. It was ugly and I don't know enough HTML/CSS/whatever to make it not-ugly.

I looked into injecting the main Kubos docs link:
- The rustdoc `--html-in-header` option sounds promising, but [isn't supported by 'cargo doc'](https://github.com/rust-lang/cargo/issues/331)
- Even if it was, our main docs link would probably refer to 'latest', so if a user was browsing a different version, they might not realize it got swapped
- We could potentially insert a 'navigate back to previous page' link, but I decided to not continue down that rabbit hole

plauche:
Honestly I would almost rather not integrate the rust docs until we can display them in a uniform fashion alongside the rest of our docs. But if we really need the rust docs for this release then this current integration is probably the bare minimum to get us there. If we go ahead with this then we should revisit rust docs integration and dig into a better integration. @marshall may also have opinions on how this looks.

CatGarab:
@plauche That's a fair point. I would certainly like the docs to play nicely together and look the same, I just don't have the knowledge required to be able to make that happen quickly. The major changes for this next release revolve around our new APIs and services, most of which are only documented through these Rust docs. I'm super against mentioning that we have support for something without any docs about the thing, so that means doing ugly links for now

PR #109, opened 2018-06-20
Increasing OEM6 receive timeout

CatGarab:
When requesting version information from the OEM6 (which is the only command that does on-demand requests for information), the current timeout when waiting for the information to be returned is slightly too slow sometimes. When I was debugging this, I saw return times of around 305ms.

Increasing the timeout so that we can get the version info successfully (consistently)


CatGarab:
[Fix OEM6 service](https://trello.com/c/OETDJt4z/375-fix-oem6-service)

PR #112, opened 2018-06-25
Added lua sdk doc and communication service doc.

plauche:


plauche:
[Release Documentation](https://trello.com/c/lXP6Xyqx/364-release-documentation)

docs/services/communication-service.rst
```diff
@@ -0,0 +1,136 @@
...
+    path = "/dev/ttyUSB1"
+    baud = 9600
+
+
+Debug Serial
+~~~~~~~~~~~~
+
+The ``debug-serial`` transport allows routing of communication data over the debug console.
+This transport is meant to be used on an embedded target with debug console and
+is primarily meant to be used when no other serial ports are available.
```
CatGarab: Is there different logic under the covers for using the debug console vs a different serial port? Linux doesn't make any distinction between them (ex. for the iOBC and BBB, the debug console is just `/dev/ttyS0`)

docs/services/communication-service.rst
```diff
@@ -0,0 +1,136 @@
...
+    - ``baud`` - The speed of serial communications
+
+Example
+
+::
+
+    [[communication-service]]
+    name = "Dev Serial"
+    transport = "serial"
+    path = "/dev/ttyUSB1"
```
CatGarab: Personal Question: Are USB ports really considered regular serial devices?
nit: If so, it would be nice if one of the examples was changed to use a UART device name (`/dev/ttyS2`)

docs/services/communication-service.rst
```diff
@@ -0,0 +1,136 @@
...
+Debug Serial
+~~~~~~~~~~~~
+
+The ``debug-serial`` transport allows routing of communication data over the debug console.
+This transport is meant to be used on an embedded target with debug console and
+is primarily meant to be used when no other serial ports are available.
+
+.. note::
+   This transport will take full control of the debug console. The only way to terminate it
+   is by terminating the service, usually by rebooting the device.
```
CatGarab: What would be other ways to terminate the service?

docs/services/core-services.rst
```diff
@@ -3,13 +3,20 @@ Core Services
...
+    :maxdepth: 1
+    :hidden:
+
+    Communication Service <communication-service>
+
 Upcoming services are:
 
  - |telem-db|
- - Communication Service
+ - Communication Service <communication-service>
```
CatGarab: This isn't "Upcoming" anymore, so can be removed from this list

docs/services/communication-service.rst
```diff
@@ -0,0 +1,136 @@
...
+======================
+
+The `communications service <https://github.com/kubos/kubos/tree/master/services/communication-service>`__
+provides a bridge for communicating with the satellite
+in both development and flight environments. The service itself is built to work
+across multiple different transports and is easily expandable to work over
+additional transports in the future. Currently the communications service is
+implemented in :doc:`Lua <../sdk-docs/sdk-lua>`, but in the future it will
+likely be implemented in Rust.
+
```
jacoffey3: Before jumping straight into running, add a section here that discusses how it is typically implemented in the system. (eg - service talks to the ends of a gateway).

docs/services/communication-service.rst
```diff
@@ -0,0 +1,136 @@
+Communications Service
+======================
+
+The `communications service <https://github.com/kubos/kubos/tree/master/services/communication-service>`__
+provides a bridge for communicating with the satellite
+in both development and flight environments. The service itself is built to work
+across multiple different transports and is easily expandable to work over
```
jacoffey3: The way you're using transports here is actually named gateways in our architecture. I suggest changing it to be gateway most of the time, and only use transport when you're describing UDP vs TCP vs serial, not the full link.

docs/services/communication-service.rst
plauche: Yea it basically hijacks the console and takes total control of stdin/out. I forgot that this transport doesn't need to know which serial port the debug console runs on, because using this transport means you are running the service directly in the debug console already.

docs/services/communication-service.rst
plauche: It is pretty normal for a usb to serial cable to be presently as `/dev/ttyUSBX`.

docs/services/communication-service.rst
plauche: There are not really other options unless you have another way to get a console on the device, in which case we wouldn't recommend taking over the debug console.

docs/services/communication.rst
```diff
@@ -0,0 +1,177 @@
...
+    baud = 115200
+
+Transports
+----------
+
+Transports are different sources of messages for the communication service. They will
+often take the form of physical layers such as serial links or radios. The communication
+service should always be configured with one UDP transport to allow communication with
+local services. A single communication service instance can only be configured
+to work with two transports at a time.
```
CatGarab: Is it up to two transports, or is exactly two transports?

docs/services/communication.rst
```diff
@@ -0,0 +1,177 @@
...
+physical transports. It is intended to always be used with one UDP transport
+and one non-UDP transport. Currently there are two serial-based transports
+available, with radio-based transports in the works.
+
+A typical production setup would involve one communication service instance
+running on the flight OBC and one instance running on the ground segment.
+Each instance would use a UDP transport in order to communicate with
+local services or clients. The OBC's instance would use a transport
+custom to the radio gateway. The ground's instance would also use
+a custom transport designed for the ground side of the gateway.
```
CatGarab: I really like these overview paragraphs :100:

PR #113, opened 2018-06-25
Added new circleci step to regenerate just the docs on a release tag.

plauche:
Our current Circle configuration will not run when release tags are generated. This requires us to manually run the docs generation step for each release. These changes should instruct Circle to just regenerate and publish the docs when a new release (`x.y.x` *not* `x.y.z+abc`) tag is made. Following configuration instructions from here - https://circleci.com/docs/1.0/configuration/#tags.


plauche:
[Generate docs on release tag](https://trello.com/c/A8v4SqpN/385-generate-docs-on-release-tag)

PR #115, opened 2018-06-26
More Circle docs fixes

plauche:
- Combine kubos-dev/doc docker containers
- Added yotta step needed for cargo to generate docs


tools/dist/Dockerfile
```diff
@@ -1,4 +1,4 @@
-# kubos/kubos-dev:0.1.2
+# kubos/kubos-dev:0.1.3
```
CatGarab: ...I totally didn't notice this field when I was making changes :sweat_smile:  The next version will actually be 0.1.6

PR #118, opened 2018-06-29
Python app api

jacoffey3:
app api only does socket management for queries and error classification.

Also contains changes to pumpkin-mcu-service for formatting.


jacoffey3:
[Integration Testing Framework](https://trello.com/c/R8vEqdkV/353-integration-testing-framework)

jacoffey3:
[Python App API](https://trello.com/c/GHDm6D6J/366-python-app-api)

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,82 @@
...
+        data = self._format(response, service)
+
+        return data
+
+    def _udp_query(self, query, (ip, port)):
+        try:
+            # Set up the socket
+            sock = socket.socket(socket.AF_INET,  # Internet
+                                 socket.SOCK_DGRAM)  # UDP
+            sock.settimeout(QUERY_TIMEOUT)
```
marshall: minor, but in my latest changes, the socket timeout is a parameter to `query` instead of statically defined:
https://github.com/kubos/kubos/pull/91/files#diff-b40535f4abd4d4722fc06900db02f7ffR171

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,82 @@
...
+QUERY_TIMEOUT = 10.0  # Seconds
+BLOCKING = 1  # 1 for blocking
+
+
+class Services:
+
+    def __init__(self, config_filepath=DEFAULT_CONFIG_PATH):
+        self.config = toml.load(config_filepath)
+
+    def query(self, service, query):
```
marshall: another minor diff with the latest API, I replaced `service` with `host_url` in the form `{IP}:{PORT}`, which offers greater flexibility for testing in the future. KubOS `Config` objects have a handy `hosturl()` function that generates this address format directly. https://github.com/kubos/kubos/pull/91/files#diff-b40535f4abd4d4722fc06900db02f7ffR171

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,81 @@
...
+            raise TypeError("Query must be str or unicode.")
+
+        # Lookup port/ip
+        ip = self.config[service]["addr"]["ip"]
+        port = self.config[service]["addr"]["port"]
+
+        # Talk to the server
+        response = self._udp_query(query, (ip, port), timeout)
+
+        # format the response and detect errors
```
CatGarab: nit: capitalize Format

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def _format(self, response, service):
+
+        # Parse JSON response
+        response = json.loads(response)
+
+        # Check that it follows GraphQL format
+        if 'msg' not in response or 'errs' not in response:
+            raise KeyError(
+                "{} Endpoint Error: ".format(service) +
```
CatGarab: Change the error description. `KeyError` is different from `EnvironmentError`, so the description should reflect that.

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,81 @@
...
+                "Response contains incorrect fields: \n{}".format(response))
+        # Parse response according to GraphQL standard format
+        data = response['msg']
+        errors = response['errs']
+
+        # Check for endpoint errors
+        if errors not in ([], None):
+            raise EnvironmentError(
```
CatGarab: Why not just name this `EndpointError`?

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,81 @@
...
+            # Wait for response (until timeout occurs)
+            response = sock.recv(UDP_BUFF)
+            return response
+        finally:
+            sock.close()
+
+    def _format(self, response, service):
+
+        # Parse JSON response
+        response = json.loads(response)
```
CatGarab: What happens if the response isn't proper JSON? (Most likely to occur if the response exceeds the UDP buffer size)

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,81 @@
...
+"""
+Mission Application API for Python Mission Applications.
+"""
+
+import toml
+import socket
+import json
+
+SERVICE_CONFIG_PATH = "/home/system/etc/config.toml"
+UDP_BUFF = 1024
```
plauche: nit: The name `UDP_BUFF_LEN` would be a bit more clear

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,81 @@
...
+            sock.bind(("", 0))  # Binds to an available port
+
+            # Send Query
+            sock.sendto(query, (ip, port))
+
+            # Wait for response (until timeout occurs)
+            response = sock.recv(UDP_BUFF)
+            return response
+        finally:
+            sock.close()
```
plauche: Do you intend for `sock.close()` to only get called when an exception is thrown? Right now it is not called when everything is successful, however the socket may still be closed when `sock` goes out of scope.

apis/python-app-api/app_api.py
plauche: Is raising an exception the appropriate behaviour here? Technically you have received a valid graphql response back. It just contains error information from the service.

apis/python-app-api/app_api.py
jacoffey3: It does, the next line describes what's happening. Both are endpoint errors, and could possibly be described as environment errors, but I decided to make the one a key error because that is what Python would raise if I didn't catch it and describe it manually.

apis/python-app-api/app_api.py
jacoffey3: It requires making a new error class in Python. I don't want to make a new error class. We will eventually need to make one most likely, but it's really not nearly as important as the message itself.

apis/python-app-api/app_api.py
jacoffey3: Is is called when everything is successful. This behaves the same as `with` for a file handle. The `finally` modifier means that it is executed _no matter what_ after the code within the try statement executes. This means that even if an exception is encountered, the socket is closed no matter what. I implemented this because I was starting to get issues with crossover between queries if I kept the same socket bound and I had an endpoint misbehaving.

apis/python-app-api/app_api.py
jacoffey3: I think so. I'm unsure if the API should focus on throwing errors to bring attention, or if it should focus on completing the transaction no matter what. It's a conversation we need to have as we start implementing mission apps. The reason I implemented it this way is because I want mission apps to have the final say in error detection and recovery. For instance, if I'm going through my mission logic, as long as the request doesn't raise an error, I know without a doubt that my data is good using this API and I can proceed without any additional logic. The downside is I have to do error catching in my mission app for when I get bad data from the endpoint. _But,_ since it's bad data anyways, I think it's better to force mission apps to deal with that rather than handle it for them, since even if I didn't raise an exception, they should be dealing with it anyways.

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def _format(self, response, service):
+
+        # Parse JSON response
+        try:
+            response = json.loads(response)
+        except Exception as error:
```
jacoffey3: @CatGarab check out what I did here. It still allows the error thrown by the library to go through, and preserves the stacktrace, but gives some more context.

apis/python-app-api/app_api.py
plauche: But this is not bad data, this is a valid response with error information. No errors have occurred in the Python code. I don't think the exception is the appropriate way to handle this situation.

This is also a pretty low level function which appears to just be extracting information from the graphql response (at least the name `format` indicates as much). I think a more appropriate response here would be to `return (data, errors)` and let the caller then correctly handle and parse the error or data. Ideally application writers won't be calling this function directly right? It seems like something which would be wrapped up in an higher level API that does the error handling and throwing if necessary.

apis/python-app-api/app_api.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # Talk to the server
+        response = self._udp_query(query, (ip, port), timeout)
+
+        # Format the response and detect errors
+        (data, errors) = self._format(response, service)
+
+        # Check for endpoint errors
+        if errors not in ([], None):
+            raise EnvironmentError(
```
jacoffey3: @plauche I moved the error creation for endpoint errors higher up here so it would not be buried in a lower level function. 

As for whether or not it _should_ throw an error, I still think it absolutely should. I agree the errors returned by the GraphQL endpoints should absolutely be preserved and handled by the user. The only difference is _how_ the user can handle it. The user is able to ignore the error field if we simply return it, and will not know there are any errors unless they check it. If we raise and error and return the error in the exception description, we are still returning the information, but forcing the user to deal with the error. This is more following a philosophy making tools to keep the user from doing bad things. I think since we are expecting users that are novice coders to use this library, I want to make sure they are forced to use it correctly.

PR #123, opened 2018-07-05
Minor docs fixes

cantino:
Here are some minor fixes to typos in the docs that I found as I was reading through them.


cantino:
Thanks!

PR #125, opened 2018-07-11
hardware services test script

jacoffey3:
Changes to the pumpkin mcu service and a hardware services testing script that uses the python app package.


jacoffey3:
[Integration Testing Framework](https://trello.com/c/R8vEqdkV/353-integration-testing-framework)

libs/kubos-test/README.md
```diff
@@ -0,0 +1,18 @@
+# Kubos Manual Testing Package
+
+This python module is meant to be used for easy manual testing of KubOS services. Currently, integration testing for hardware services is implemented. This utilizes the "testHardware" mutation for each hardware service outline in the hardware service schema.
```
CatGarab: nit: Capitalize Python :stuck_out_tongue:

libs/kubos-test/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Usage
+
+An example is provided for creating an integration test script in the code below.
+
+.. code:
```
CatGarab: Convert to markdown formatting (triple tics)

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python
+
+# Copyright 2018 Kubos Corporation
+# Licensed under the Apache License, Version 2.0
+# See LICENSE file for details.
+
+"""
+Unit testing for the I2C library.
```
CatGarab: nit: Update comment

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+import app_api
+import socket
+
+DEFAULT_CONFIG_PATH = "/home/system/etc/config.toml"
+SERVICE_MUTATION = (
+    'mutation {testHardware(test:"Integration"){errors,status}}')
```
CatGarab: 1. (Atleast in Rust) Enums are give in all-caps, without quotes (so `test: INTEGRATION`)
2. [Per the service outline doc...](http://docs.kubos.co/1.3.0/dev-docs/service-outline-guide.html) status -> success

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,56 @@
...
+        for service in self.api.config:
+            try:
+                response = self.api.query(
+                    service=service,
+                    query=SERVICE_MUTATION,
+                    timeout=QUERY_TIMEOUT)
+                print "Status: SUCCESS\n {}".format(service)
+                print "Response: {}\n".format(response)
+            except socket.timeout as e:
+                print "Status: FAILED\n {}".format(service)
```
CatGarab: Could change this to "Status: TIMEDOUT", removing the need for the other two messages

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_services(self):
+        for service in self.api.config:
+            try:
+                response = self.api.query(
+                    service=service,
+                    query=SERVICE_MUTATION,
+                    timeout=QUERY_TIMEOUT)
+                print "Status: SUCCESS\n {}".format(service)
+                print "Response: {}\n".format(response)
+            except socket.timeout as e:
```
CatGarab: :+1: Great error handling

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,56 @@
...
+                print "Status: FAILED\n {}".format(service)
+                print "Error: No response from server"
+                print " Timeout: {} seconds\n".format(QUERY_TIMEOUT)
+            except Exception as e:
+                print "Status: FAILED\n {}".format(service)
+                print "Error: {}, {}\n".format(type(e), e)
+
+    def test_service(self, service):
```
CatGarab: What's this function for? It doesn't appear to be called by anything.

If you're leaving the possibility that someone might want to only test a single service at some point, change the `for` loop in `test_services` to call this instead.

So (warning - horrible psuedocode):
```
def test_services(self, service=None):
    if service != None:
        test_service(service)
    else:
        for elem in list-o-services:
            test_service(elem)
```

libs/kubos-test/setup.py
```diff
@@ -0,0 +1,14 @@
...
+See:
+https://github.com/pypa/sampleproject
+"""
+
+from setuptools import setup
+
+setup(name='kubos_test',
+      version='0.1.0',
+      description='Manual integration testing library for KubOS Services',
+      py_modules=["kubos_test"]
```
CatGarab: include the package dependencies

services/pumpkin-mcu-service/service/schema.py
```diff
@@ -119,12 +119,47 @@ def mutate(self, info, module, command):
         return commandStatus
 
 
+class TestHardware(graphene.Mutation):
+    """
+    Tests if the hardware is present and talking.
+    """
+
+    class Arguments:
+        test = graphene.String()
```
CatGarab: Test type should be an [enum](http://docs.graphene-python.org/en/latest/types/enums/), rather than a string

services/pumpkin-mcu-service/service/schema.py
```diff
@@ -119,12 +119,47 @@ def mutate(self, info, module, command):
...
+    def mutate(self, info, test):
+
+        test_output = {}
+        status = True
+        errors = None
+        if test == "Integration":
+            for module in MODULES:
+                mcu = mcu_api.MCU(address=MODULES[module]['address'])
+                out = mcu.read_telemetry(module=module, fields=['all'])
+                test_output.update(out)
```
CatGarab: You need to handle what happens when one of the devices is offline

libs/kubos-test/kubos_test.py
jacoffey3: Ohh this might be a problem because python doesn't like that

services/pumpkin-mcu-service/service/schema.py
jacoffey3: It errors, as desired

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+Currently only implements integration testing for hardware services.
+"""
+
+import app_api
+import socket
+
+DEFAULT_CONFIG_PATH = "/home/system/etc/config.toml"
+SERVICE_MUTATION = (
+    'mutation {testHardware(test:INTEGRATION){errors,status}}')
```
CatGarab: Github marked my old comment on this as outdated...
The service outline doc and the current rust services use "success" instead of "status". You said that python has issues with that. We need to figure out how to commonize it, then.

libs/kubos-test/kubos_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class IntegrationTest:
+
+    def __init__(self,
+                 config_filepath=DEFAULT_CONFIG_PATH,
+                 output_file=None):
+        self.api = app_api.Services(config_filepath)
+        if output_file is not None and type(output_file) == str:
+            self.logfile = output_file
```
CatGarab: `logfile` doesn't appear to be used anywhere

libs/kubos-test/kubos_test.py
jacoffey3: Your original comment was on test type. I just didn't notice the success bit. I can change that.

libs/kubos-test/kubos_test.py
jacoffey3: Yes I forgot I was still implementing that...I'm working on giving the option to dump stdout into a logfile, per your previous comments

jacoffey3: So I poked around, and you can easily redirect the output whenever you actually run the test. In Python, you can just run `python integration_test.py > filename` and it will redirect the output to the file. I could add this capability to my library, but it's so simple it seems unnecessary.

PR #126, opened 2018-07-11
[WIP] Timestamps should be milliseconds

cantino:
Update to use `DOUBLE` for the `timestamp` column in sqlite and `f64` in Rust.

@cantino paired with @plauche.


CatGarab:
This work has been completed in other PRs. Closing.

PR #137, opened 2018-07-24
Remove debug prints from Telemetry DB API

marshall:
This causes a metadata parsing issue for KubOS apps that use the telemetry DB. The long term fix for this problem is moving to standalone metadata in an app package (https://trello.com/c/QNrjsmh8/421-move-metadata-out-of-app-binary-and-into-an-app-archive-ie-signed-zip)


plauche:
I rebased this branch with master and found the change was already there!

PR #139, opened 2018-07-25
Migrating app metadata to external file source

CatGarab:
When registering a new application, the app service currently calls the application with the "-m" flag to get the metadata information.
The code to print said-metadata is language specific.
This PR updates the registration process to read the metadata from a manifest file instead (`manifest.toml`), reducing the amount of code required when writing an application in a new language.

Ex:

```
name = "myapp"
version = "1.0"
author = "me"
```


CatGarab:
[Move metadata out of app binary](https://trello.com/c/QNrjsmh8/421-move-metadata-out-of-app-binary)

PR #140, opened 2018-07-27
Commonizing Python and Rust app APIs

CatGarab:
We want the user experience with the Python and Rust app APIs to be as similar as possible.

In this PR:
- Reorged the api folders to be `rust` and `python` under one parent directory `app-api`
- Changed the Rust APIs `query` function to take the service name to query instead of the IP address
- Changed the Rust APIs behavior when an unknown run level is passed. Instead of executing the `OnCommand` run level, it will now print an error and exit


jacoffey3:
Need to change the circle config for the new API location to run the tests

apis/app-api/rust/src/query.rs
```diff
@@ -47,7 +51,7 @@ type AppResult<T> = Result<T, failure::Error>;
 /// 		ping
 /// 	}"#;
 ///
-/// let result = query("0.0.0.0:1234", request, Some(Duration::from_secs(1)))?;
+/// let result = query("radio-service", Some("/home/kubos/config.toml"), request, Some(Duration::from_secs(1)))?;
```
marshall: this seems a lot less flexible than the host/port version of the method. previously, to pull the host/port from Config a user just needed to use the `Config::new` or `Config::new_from_path` constructor along with `config.hosturl()`. Now there is no way to query a service without it being registered in a config file somewhere, which seems bad for for both testing, and auto-discovery.

tl;dr just for comparison sake, it seems like we're making this API harder than it needs to be, see my comparison of old vs new below.

```rust
let timeout = Some(Duration::from_secs(1));
let request = "{ .. }";

// old
query(Config::new("radio-service").hosturl(), request, timeout);
// vs. new
query("radio-service", None, request, timeout);

// old
query(Config::new_from_path("radio-service", "/home/kubos/config.toml").hosturl(), request, timeout);
// vs. new
query("radio-service", Some("/home/kubos/config.toml"), request, timeout);
```

I admit the "old" API is a bit more verbose, but I think that could also be easily fixed with a `impl FromStr for Config`

marshall: btw, if you feel it's important to abstract away the host/port aspect of the old API, maybe we just pass the `Config` ref directly as one argument, and let the various `Config` constructors cover the bases of how a Config is constructed, i.e:

```rust
pub fn query(config: &Config, request: &str, timeout: Duration) ...
```

this would make the above examples look like:
```rust
query(Config::new("radio-service"), request, timeout);
query(Config::new_from_path("radio-service, "/home/kubos/config.toml"), request, timeout);
```

in this world, we could also add more helper `Config` constructors for testing/auto-discovery purposes, and not have to make different versions of `query`

CatGarab: -  Honestly never occurred to me that I could do `query(Config::new("radio-service").hosturl(), request, timeout);` instead of having to manually look up the IP and port for the service. So if this code gets changed back, that'll absolutely need to get documented as a possibility.
-  > Now there is no way to query a service without it being registered in a config file somewhere, which seems bad for for both testing, and auto-discovery.
    - I think that requiring all `query()` calls to go against a known/configured service is good. I'm not sure why it would be bad for testing. We have no concept of auto-discovery.
- `pub fn query(config: &Config, request: &str, timeout: Duration) ...` :+1: I like that. It's neater than what I did.

PR #146, opened 2018-08-01
Allowing `"Errs": null` in GraphQL Response

CatGarab:
Currently, when the Rust `kubos-service` crate processes a good graphql request, it returns `{"errs": "", "msg": {response data}}`, while the Python version returns `{"errs": null, "msg": {response}}`.

This PR updates the Rust `query()` function to allow either format.


CatGarab:
[Telemetry collection and beacon app](https://trello.com/c/6pVP00MQ/398-telemetry-collection-and-beacon-app)

PR #153, opened 2018-08-23
File protocol reorg

plauche:


libs/file-protocol/src/protocol.rs
```diff
@@ -0,0 +1,349 @@
...
+                    if ack_hash == hash_val {
+                        // Done processing its time to go home
+                        // return Ok(true);
+                    }
+                }
+                // return Ok(true);
+            }
+            Ok(Message::NAK(hash, Some(missing_chunks))) => {
+                info!("<- {{ {}, false, {:?} }}", hash, missing_chunks);
+                if let Some(chunks) = Some(missing_chunks) {
```
CatGarab: Since you've got `Some(missing_chunks)` in the match case, you don't need to test for anything here. You can just do `self.do_upload(&hash, &missing_chunks)`

libs/file-protocol/src/protocol.rs
```diff
@@ -0,0 +1,349 @@
...
+                    self.do_upload(&hash, &chunks)?;
+                }
+            }
+            Ok(Message::NAK(hash, None)) => {
+                info!("<- {{ {}, false }}", hash);
+            }
+            Ok(Message::ReqReceive(channel_id, hash, path, Some(mode))) => {
```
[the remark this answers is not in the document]
plauche: It is possible that we fail to read the mode metadata so we don't send a mode across the wire. In the lua code this resulted in using a default mode on the receiving end. We could just as easily put the default in the sender and make the mode non-optional.

libs/file-protocol/src/protocol.rs
```diff
@@ -0,0 +1,318 @@
...
+                            channel_id, hash, path, mode
+                        );
+                        // The client wants to send us a file.
+                        // Go listen for the chunks.
+                        // Note: Won't return until we've received all of them.
+                        // (so could potentially never return)
+                        // TODO: handle channel_id mismatch
+                        self.send(messages::ack_or_nak(&hash, None).unwrap())
+                            .unwrap();
+                        new_state = State::Receiving(
```
CatGarab: If you send an ACK, that means you already have all the chunks you need, so you don't need to receive anything. In that case, state should become `ReceivingDone`

libs/file-protocol/src/storage.rs
```diff
@@ -0,0 +1,312 @@
...
+            }
+        }
+    }
+    store_meta(&hash, index).unwrap();
+
+    if let Ok(meta) = fs::metadata(source_path) {
+        Ok((hash, index, meta.mode()))
+    } else {
+        // default mode to 0o644
+        Ok((hash, index, 0x1a4))
```
CatGarab: Does Rust not support octals?

services/file-service-rust/src/main.rs
```diff
@@ -0,0 +1,42 @@
...
+extern crate log;
+extern crate simplelog;
+
+use file_service_rust::*;
+use kubos_system::Config as ServiceConfig;
+use simplelog::*;
+
+fn main() {
+    CombinedLogger::init(vec![
+        TermLogger::new(LevelFilter::Info, Config::default()).unwrap(),
```
CatGarab: This needs to get changed/handled. If the service is started automatically at boot time, there's no terminal to bind to, so `TermLogger::new()` fails. As a result, the service panics.

libs/file-protocol/src/protocol.rs
```diff
@@ -0,0 +1,339 @@
...
+                            "Transmission failure on channel {}. Error returned from server: {}",
+                            channel_id, error_message
+                        ));
+                    }
+                }
+                Ok(new_state)
+            }
+            Err(e) => {
+                info!("<- what did we get?? {}", e);
+                match state.clone() {
```
CatGarab: You've got basically this logic in both of the places where this error would be detected, so this section could be removed. (Maybe keep the `State::Holding` though...)

PR #158, opened 2018-09-05
Deployment doc

jacoffey3:


plauche:
Do you think it is necessary to mention that having an accurate on-board RTC is crucial for making this work?

jacoffey3:
@plauche added a note.

PR #164, opened 2018-09-07
Removing lua file stuff

plauche:
Deletes lua file service, client, protocol lib. Renames rust pieces to match old names.


plauche:
[Rust File Transfer Service](https://trello.com/c/KGpwS3xp/46-rust-file-transfer-service)

PR #166, opened 2018-09-10
Adding intra-transaction message timeout config option

CatGarab:


CatGarab:
[File service: timeout config option](https://trello.com/c/Zxolh7r7/63-file-service-timeout-config-option)

PR #168, opened 2018-09-11
Adding `insert` mutation to telemetry DB service

CatGarab:
Other changes:

- Modifying test utils to allow the database path and service port number to be configurable on a per-test basis. This allows us to run tests simultaneously
- Updating `test_filter_timestamp` to break test cases into their own tests


CatGarab:
[Telemetry DB: mutation to write to the database](https://trello.com/c/I1pON53F/72-telemetry-db-mutation-to-write-to-the-database)

PR #169, opened 2018-09-12
Adding query to allow telemetry database results to be routed to an output file

CatGarab:


CatGarab:
[Telemetry DB: Optional query argument with filename to write response to a compressed file](https://trello.com/c/kowbWLWf/73-telemetry-db-optional-query-argument-with-filename-to-write-response-to-a-compressed-file)

docs/services/telemetry-db.rst
```diff
@@ -43,6 +43,32 @@ Each of the query arguments acts as a filter for the database query:
 Note: ``timestampGe`` and ``timestampLe`` can be combined to create a timestamp selection range.
 For example, entries with timestamps after ``1000``, but before ``5000``.
 
+Saving Results for Later Processing
+-----------------------------------
+
+Sometimes we might want to periodically query the database for telemetry data, but not immediately use the results.
```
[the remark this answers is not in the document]
jacoffey3: This way we can use the File Service [link] to transfer the file as the space ground connection allows.

services/telemetry-service/src/schema.rs
```diff
@@ -56,42 +110,53 @@ graphql_object!(QueryRoot: Context |&self| {
...
+        let file_name = file_name_raw.to_str().ok_or_else(|| return FieldError::new("Unable to parse output file name to string", Value::null()))?;
+        
+        if let Some(parent) = output_path.parent() {
+            fs::create_dir_all(parent)?;
         }
-
-        if let Some(l) = limit {
-            query = query.limit(l.into());
+        
+        {
```
plauche: nit: these `{}`s look like they aren't needed

CatGarab: They're actually on purpose. It makes the `File` go out of scope, so everything gets officially written and synced with the file system.

I found that if I didn't do it this way, the compression logic would basically just read an empty file.

PR #170, opened 2018-09-13
Telemetry  DB service direct UDP connection

CatGarab:


CatGarab:
[Telemetry DB: UDP socket to write to the database asynchronously](https://trello.com/c/kiL1NIWx/74-telemetry-db-udp-socket-to-write-to-the-database-asynchronously)

jacoffey3:
@plauche I think it might be a bit much to ask at this point. Also @CatGarab, are the entries worked off round-robin from the 2 different sockets, or does one take priority?

CatGarab:
@jacoffey3 it's not an official round-robin; it's two threads of the same priority level with a mutex, so it's just whichever thread gets triggered first (in theory, it'll go back and forth, though).

jacoffey3:
> @jacoffey3 it's not an official round-robin; it's two threads of the same priority level with a mutex, so it's just whichever thread gets triggered first (in theory, it'll go back and forth, though).

That should be fine then! That'll keep anything from getting locked out.

PR #171, opened 2018-09-13
File Service Threading Update

plauche:
The threading model of the file service has been updated to better support multiple simultaneous clients uploading or downloading files. This has also relieved the need for most non-timeout related delays in the protocol implementation :tada:.

Other majorish changes:
- Content addressable messages (metadata, chunk, ack, nak) now include the channel id as first parameter
- Channel ids moved from `u64` -> `u32`

### DIFF
--- a/docs/services/file-protocol.rst
+++ b/docs/services/file-protocol.rst
@@ -66,9 +66,8 @@ Messages
 All messages in the file protocol are encoded as `CBOR <http://cbor.io/>`__ arrays and are sent
 in UDP packets.
 
-The first value in the encoded list is the ``channel_id``
-for request/response type messages and it is followed by
-the ``hash`` for content-addressable messages.
+The first value in the encoded list is the ``channel_id`` for all messages
+and it is followed by the ``hash`` for content-addressable messages.
 
     - The ``channel_id`` parameter is used to indicate a group of messages associated with
       a particular file protocol transaction.
--- a/docs/services/file-protocol.rst
+++ b/docs/services/file-protocol.rst
@@ -67,33 +67,33 @@ All messages in the file protocol are encoded as `CBOR <http://cbor.io/>`__ arra
 in UDP packets.
 
 The first value in the encoded list is the ``channel_id``
-for request/response type messages or the ``hash`` for content-addressable
-messages.
+for request/response type messages and it is followed by
+the ``hash`` for content-addressable messages.
 
     - The ``channel_id`` parameter is used to indicate a group of messages associated with
       a particular file protocol transaction.
     - The ``hash`` parameter is the BLAKE2 hash for the corresponding file
       which is being transferred.
 
-+-------------------------------+----------------------------------------------------------------+
-| Name                          | Syntax                                                         |
-+===============================+================================================================+
-| `Metadata`_                   | { `hash`, `num_chunks` }                                       |
-+-------------------------------+----------------------------------------------------------------+
-| `Export Request`_             | { `channel_id`, export, `hash`, `path`, `mode` }               |
-+-------------------------------+----------------------------------------------------------------+
-| `Import Request`_             | { `channel_id`, import, `path` }                               |
-+-------------------------------+----------------------------------------------------------------+
-| `File Chunk`_                 | { `hash`, `chunk_index`, `data` }                              |
-+-------------------------------+----------------------------------------------------------------+
-| `Acknowledge (ACK)`_          | { `hash`, true, `num_chunks` }                                 |
-+-------------------------------+----------------------------------------------------------------+
-| `Negative Acknowledge (NAK)`_ | { `hash`, false, `x_start`, `x_end`, `y_start`, `y_end`, ... } |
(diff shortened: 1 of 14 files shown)


services/file-service/src/lib.rs
```diff
@@ -59,30 +65,56 @@ pub fn recv_loop(config: ServiceConfig) -> Result<(), String> {
...
-        // Break the processing work off into its own thread so we can
-        // listen for requests from other clients
-        thread::spawn(move || {
-            let mut state = State::Holding {
-                count: 0,
-                prev_state: Box::new(State::Done),
-            };
+        let channel_id = match file_protocol::parse_channel_id(&first_message) {
+            Ok(Some(channel_id)) => channel_id,
+            Ok(None) => 0,
```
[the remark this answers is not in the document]
plauche: Ahh I need to clean that up. Originally I was working around the messages without channel IDs, but this code is no longer needed.

services/file-service/src/lib.rs
```diff
@@ -59,30 +65,56 @@ pub fn recv_loop(config: ServiceConfig) -> Result<(), String> {
...
-            if let Some(msg) = first_message {
-                if let Ok(new_state) = f_protocol.process_message(msg, state.clone()) {
-                    state = new_state;
+                // Listen, process, and react to the remaining messages in the
+                // requested operation
+                match f_protocol.message_engine(
+                    |d| match receiver.recv_timeout(d) {
+                        Ok(v) => Ok(v),
+                        Err(RecvTimeoutError::Timeout) => Ok(None),
+                        Err(e) => Err(Some(format!("Error {:?}", e))),
```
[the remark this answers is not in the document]
plauche: Hmm do I still need this for propagating other non-timeout errors?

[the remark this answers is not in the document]
CatGarab: Oh, I phrased that poorly. Take out the `Some` and just do `Err(format!("Error {:?}", e)),` instead

docs/services/file-protocol.rst
```diff
@@ -67,33 +67,33 @@ All messages in the file protocol are encoded as `CBOR <http://cbor.io/>`__ arra
 in UDP packets.
 
 The first value in the encoded list is the ``channel_id``
-for request/response type messages or the ``hash`` for content-addressable
-messages.
+for request/response type messages and it is followed by
```
CatGarab: nit: You no longer need "for request/response type messages" since `channel_id` is now a part of all messages

libs/cbor-protocol/src/lib.rs
```diff
@@ -396,12 +399,14 @@ impl Protocol {
         let (size, _peer) = match result {
             Ok(data) => data,
             Err(err) => match err.kind() {
-                ::std::io::ErrorKind::WouldBlock => return Err(None), // For some reason, UDP recv returns WouldBlock for timeouts
-                _ => return Err(Some(format!("Failed to receive a message: {:?}", err))),
+                ::std::io::ErrorKind::WouldBlock => {
```
CatGarab: This is the timeout error, so this should return `Ok(None)`

libs/file-protocol/src/parsers.rs
```diff
@@ -327,29 +307,26 @@ pub fn parse_chunk(data: Vec<Value>) -> Result<Option<Message>, String> {
...
+                    return Ok(Some(Message::Metadata(
+                        channel_id,
+                        hash.to_owned(),
+                        *num as u32,
+                    )));
                 }
             }
         } else {
             // It's a sync message: { hash }
             // TODO: Whoever processes this message should do the sync_and_send
```
CatGarab: nit: this todo comment should get removed

libs/file-protocol/src/protocol.rs
```diff
@@ -203,25 +204,45 @@ impl Protocol {
     ///
     /// let message = match f_protocol.recv(Some(Duration::from_secs(1))) {
     /// 	Ok(data) => data,
-    /// 	Err(None) => {
```
CatGarab: It should now be `Ok(None)` instead of `Err(None)`, but the timeout case is still a valid thing to check for here

libs/file-protocol/src/protocol.rs
```diff
@@ -203,25 +204,45 @@ impl Protocol {
...
     ///
-    pub fn recv(&self, timeout: Option<Duration>) -> Result<Option<Value>, Option<String>> {
+    pub fn recv(&self, timeout: Option<Duration>) -> Result<Option<Value>, String> {
         match timeout {
             Some(value) => self.cbor_proto.recv_message_timeout(value),
             None => self.cbor_proto.recv_message(),
         }
     }
 
+    /// Generates a new random channel id for use when initiating a
```
CatGarab: nit: capitalize ID

libs/cbor-protocol/src/lib.rs
plauche: Ahh yea good catch

PR #176, opened 2018-09-24
Adding missing service documentation

CatGarab:
I previously exposed the `objects.rs` modules of several hardware services as public so that the documentation would be generated and available. This PR documents all of the remaining fields that I missed.

Also updating rustbuild.sh to exit on error


CatGarab:
[Fix CI test not failing](https://trello.com/c/tQA2GcrU/89-fix-ci-test-not-failing)

PR #179, opened 2018-09-25
File protocol clean up

plauche:
- Clean up after a corrupt chunk
- Clean up after a hash mismatch

### DIFF
--- a/libs/file-protocol/src/protocol.rs
+++ b/libs/file-protocol/src/protocol.rs
@@ -411,8 +411,13 @@ impl Protocol {
     ) -> Result<(), ProtocolError> {
         for (first, last) in chunks {
             for chunk_index in *first..*last {
-                let chunk = storage::load_chunk(&self.prefix, hash, chunk_index)?;
-                self.send(messages::chunk(channel_id, hash, chunk_index, &chunk)?)?;
+                match storage::load_chunk(&self.prefix, hash, chunk_index) {
+                    Ok(c) => self.send(messages::chunk(channel_id, hash, chunk_index, &c)?)?,
+                    Err(e) => {
+                        warn!("Failed to load chunk {}:{} : {}", hash, chunk_index, e);
+                        storage::delete_chunk(&self.prefix, hash, chunk_index)?;
+                    }
+                };
 
                 thread::sleep(Duration::from_millis(1));
             }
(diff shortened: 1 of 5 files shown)


plauche:
[File service: Handle hash mismatch](https://trello.com/c/CQUpSWgL/61-file-service-handle-hash-mismatch)

plauche:
[FIle service: Handle corrupted chunk](https://trello.com/c/Lo3VLTI6/60-file-service-handle-corrupted-chunk)

libs/file-protocol/src/protocol.rs
```diff
@@ -410,8 +410,13 @@ impl Protocol {
     ) -> Result<(), ProtocolError> {
         for (first, last) in chunks {
             for chunk_index in *first..*last {
-                let chunk = storage::load_chunk(&self.prefix, hash, chunk_index)?;
-                self.send(messages::chunk(channel_id, hash, chunk_index, &chunk)?)?;
+                match storage::load_chunk(&self.prefix, hash, chunk_index) {
+                    Ok(c) => self.send(messages::chunk(channel_id, hash, chunk_index, &c)?)?,
+                    Err(e) => {
+                        warn!("Failed to load chunk {}:{} : {}", hash, chunk_index, e);
+                        storage::delete_chunk(&self.prefix, hash, chunk_index)?;
```
plauche: @CatGarab @jacoffey3 I'm not sure what behaviour we want in this case. We are sending a file and we fail to read one of the chunks from the file system.  The two main paths of recourse are:

- Delete the chunk, keep sending the rest of the chunks, attempt to get the chunk later (when its re-requested?)
   - We'll need to create a mechanism for storing a specific file chunk

- Delete the whole file, abort sending, send an error to the receiver, re-chunk the whole file later
   - This would be slow if the file in question is large

jacoffey3: Delete the whole file, abort sending, and send an error. Recovering this would be very difficult, and unless we can recover we should just error it out and notify about it. Put a card in the icebox about this though, because we could add it in at a later date. For now it's not necessary.

libs/file-protocol/src/storage.rs
```diff
@@ -380,7 +379,17 @@ pub fn finalize_file(
...
-        let chunk = load_chunk(prefix, hash, chunk_num)?;
+        let chunk = match load_chunk(prefix, hash, chunk_num) {
+            Ok(c) => c,
+            Err(e) => {
+                warn!(
+                    "Error encountered loading chunk {}, deleting : {}",
+                    chunk_num, e
+                );
+                delete_chunk(prefix, hash, chunk_num)?;
+                return Err(e);
```
CatGarab: Would it be worth it to let the loop attempt to load all of the chunks? That way we get all corrupted chunks in one go

PR #177, opened 2018-09-24
Adding telemetry DB service benchmarking tests

CatGarab:

PR #180, opened 2018-09-26
Adding cross-compile-friendly telemetry DB benchmark tests

CatGarab:
After finishing #177 and being asked for more data, I wanted to create a benchmark test which I could easily compile and run on an actual board.

I then immediately decided that I liked this solution far better then what I'd originally done in #177. So I'd like to nuke that PR and merge this one instead :smile:


CatGarab:
[Telemetry DB: Benchmark](https://trello.com/c/HpIbbwXK/83-telemetry-db-benchmark)

docs/services/telemetry-db.rst
```diff
@@ -16,6 +16,21 @@ Specific details about the available GraphQL queries can be found in the |telem-
...
+which we have used to measure various behaviors of the telemetry database service.
+
+Because each OBC has its own unique system resources, we recommend :doc:`compiling and running <../sdk-docs/sdk-rust>`
+the test project on your OBC to obtain the most accurate results.
+
+When run on a Beaglebone Black, we gathered the following benchmark statistics:
+
+- Sending UDP requests takes ~45 microseconds
+- Round-trip GraphQL transactions (including database insert action) take ~17 milliseconds
+- Telemetry database inserts take ~16 milliseconds
```
jacoffey3: We're going to get the question of how many samples you can insert using this. I would recommend adding the benchmarks from that perspective.

jacoffey3: Currently, I would not know how to calculate that at a glance with the listed numbers

jacoffey3: (Without the context that I have)

PR #181, opened 2018-09-27
Making the file protocol more configurable

plauche:
- [x] Chunk size
- [x] Hold state timeout

### DIFF
--- a/clients/file-client/src/main.rs
+++ b/clients/file-client/src/main.rs
@@ -7,7 +7,7 @@ extern crate failure;
 extern crate simplelog;
 
 use clap::{App, Arg};
-use file_protocol::{FileProtocol, State};
+use file_protocol::{FileProtocol, FileProtocolConfig, State};
 use simplelog::*;
 use std::path::Path;
 use std::time::Duration;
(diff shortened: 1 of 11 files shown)


plauche:
[File service: timeout-retry config option](https://trello.com/c/EytMbiVI/64-file-service-timeout-retry-config-option)

clients/file-client/src/main.rs
```diff
@@ -18,8 +18,11 @@ fn upload(
     source_path: &str,
     target_path: &str,
     prefix: Option<String>,
+    chunk_size: u32,
```
CatGarab: nit: why not just make this `usize`, if you're just going to immediately convert it?

clients/file-client/src/main.rs
```diff
@@ -100,28 +106,39 @@ fn main() {
...
+                .short("-s")
+                .takes_value(true)
+                .default_value("file-storage"),
+        ).arg(
+            Arg::with_name("chunk_size")
+                .short("-c")
+                .takes_value(true)
+                .default_value("4096"),
+        ).arg(
+            Arg::with_name("hold_timeout")
```
CatGarab: I'm not in love with calling this hold "timeout", since there is an actual time-related timeout parameter (currently hardcoded to `Duration::from_secs(2)`) which we will also want to make configurable. Maybe "hold count" instead? or "hold limit"?

PR #183, opened 2018-10-01
Fixing bug where file service crashes on invalid network data.

plauche:


jacoffey3:
[File service crashes when given bad input](https://trello.com/c/pQIzr4Sp/100-file-service-crashes-when-given-bad-input)

PR #184, opened 2018-10-01
Fixing file transfer service logging location

CatGarab:
Also tweaking telemetry db service direct udp unit tests to extend delay between the test and the verification steps


CatGarab:
[Fix file service logging](https://trello.com/c/J6WaW5vV/99-fix-file-service-logging)

PR #185, opened 2018-10-01
Updating first project tutorial

CatGarab:
This PR is the first step towards updating our new user experience with our docs.

Main changes:
- Re-writing the first project to a) be in Python and b) more explicitly go through the process to get a project running on a target OBC
- Re-structuring the first-time users landing page to support all of the tutorials that are coming

Minor changes:
- There's been a slight re-org with a new tutorials folder. Updating all the links to the old "first project" doc.


CatGarab:
[Update "First KubOS project" doc](https://trello.com/c/F7OPHTh3/102-update-first-kubos-project-doc)

docs/tutorials/first-project.rst
```diff
@@ -0,0 +1,95 @@
+Getting Started with KubOS and the Kubos SDK
+============================================
+
+This tutorial guides the user through the process of creating a new Kubos project using Python2.7
+and installing and running it on a target OBC.
+
+.. note:: 
+
+    The iOBC does not support Python. If this is the board which you are using,
```
jacoffey3: I would add a blurb about the fact that the other sections of the tutorial will still be very helpful, as many of them are not python-specific.

docs/tutorials/first-project.rst
```diff
@@ -0,0 +1,95 @@
...
+the target IP's authenticity. Enter "yes" if this occurs::
+
+    $ ssh root@10.0.2.20
+    The authenticity of host '10.0.2.20 (10.0.2.20)' can't be established.
+    ECDSA key fingerprint is SHA256:ir2TC+iML+MJ5Cb3cxTReWI69aX6EtPysFQzWleKc+8.
+    Are you sure you want to continue connecting (yes/no)? yes
+    Warning: Permanently added '10.0.2.20' (ECDSA) to the list of known hosts.
+    kubos@10.0.2.20's password: ********
+
+Please confirm that you are able to connect to the board via SSH from the SDK before proceeding
```
jacoffey3: Do we have a helper doc if you can't? I remember something like that

docs/tutorials/first-project.rst
CatGarab: Many of them are not python-specific, but they do require ethernet capabilities, which the iOBC also doesn't have :P

jacoffey3: Yes, we do have the communication service to route those IP packets and can SSH in over the serial port though. I don't want the scenario where someone using the iOBC reaches the end of the Rust doc and is just like...well...what now? without looking at the other tutorials.

jacoffey3: I'm not saying that this should say that _this exact process can be run with Rust_. I'm saying that the blurb should say that "Many of the service interactions are identical in Rust, so this guide can still be helpful if you are working in Rust on the iOBC. A Rust-focused guide for the iOBC will be added in a future release."

PR #186, opened 2018-10-03
Updating Python app API to allow response errs field to be empty

CatGarab:


CatGarab:
[Fix Python app API](https://trello.com/c/vrsQYhcW/118-fix-python-app-api)

PR #188, opened 2018-10-04
Adding overview doc for monitor service

CatGarab:


CatGarab:
[Document the monitor service](https://trello.com/c/Mh4zhagP/85-document-the-monitor-service)

docs/services/index.rst
```diff
@@ -8,6 +8,7 @@ Kubos services are defined as any persistent process that is used to interact wi
     :hidden:
 
     Core Services <core-services>
+    Monitor Service <monitor-service>
```
plauche: Should we start a new category for other services which aren't core or hardware? Or does this one belong in hardware since it is a general system hardware monitoring service?

CatGarab: Yeah, I wasn't really sure what to do with it. @jacoffey3 said that it's technically a hardware service, but it didn't really fit in with the current hardware services page

jacoffey3: It's hardware. We don't need another service category. It just happens to be the only hardware service that can run on any hardware.

PR #195, opened 2018-10-10
Add stdout/stderr handling to shell protocol

plauche:
Adds
- `stdout` message
- `stderr` message
- `exit` message
- `pid` message

Split out process handling/wrapping code into `ProcessHandler`


plauche:
[Shell protocol: read from process](https://trello.com/c/WQ8TD6SH/112-shell-protocol-read-from-process)

libs/shell-protocol/src/messages/mod.rs
```diff
@@ -32,13 +42,32 @@ pub enum Message {
...
 
 pub fn parse_message(message: ChannelMessage) -> Result<Message, ProtocolError> {
     if message.name == "spawn" {
         Ok(spawn::from_cbor(&message)?)
+    } else if message.name == "stdout" {
+        Ok(stdout::from_cbor(&message)?)
+    } else if message.name == "stderr" {
+        Ok(stderr::from_cbor(&message)?)
+    } else if message.name == "pid" {
+        Ok(pid::from_cbor(&message)?)
```
CatGarab: Missing `exit` message

libs/shell-protocol/src/messages/stderr.rs
```diff
@@ -0,0 +1,76 @@
...
+    };
+
+    Ok(Message::Stderr {
+        channel_id: message.channel_id,
+        data: data.to_owned(),
+    })
+}
+
+/// Stderr -> CBOR
+pub fn to_cbor(channel_id: u32, data: Option<&str>) -> Result<Vec<u8>, ProtocolError> {
```
CatGarab: What's the purpose or implication of having an empty stderr/stdout message? (data = None)

libs/shell-protocol/src/messages/stderr.rs
```diff
@@ -0,0 +1,76 @@
...
+        })?,
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use channel_protocol::parse_message;
+    use serde_cbor::de;
+
```
CatGarab: nit: Add a test case for empty data

libs/shell-protocol/src/process.rs
```diff
@@ -0,0 +1,132 @@
...
+                            err,
+                        })
+                    }
+                }
+            }
+            None => Ok(None),
+        }
+    }
+
+    /// Retrieve id of process
```
CatGarab: nit: capitalize ID

libs/shell-protocol/src/protocol.rs
```diff
@@ -43,27 +48,58 @@ impl Protocol {
...
+    pub fn message_engine<F>(&mut self, pump: F, timeout: Duration) -> Result<(), ProtocolError>
     where
         F: Fn(Duration) -> Result<ChannelMessage, ProtocolError>,
     {
         loop {
+            // Check child for new stdout data
+            // Reading from stdout/stderr is currently a blocking operation...
+            // This code should probably get refactored into something akin
+            // to an event loop with non-blocking io.
+            if let Some(process) = self.process.as_mut().as_mut() {
```
CatGarab: Why do you need two `as_mut()` calls?

libs/shell-protocol/src/process.rs
```diff
@@ -0,0 +1,132 @@
...
+    ///
+    /// This will block as long as the process is running.
+    /// A return value of `None` indicates the stream is
+    /// no longer available and likewise the process
+    /// is likely no longer alive.
+    pub fn read_stdout(&mut self) -> Result<Option<String>, ProtocolError> {
+        match self.stdout_reader {
+            Some(ref mut stdout_reader) => {
+                let mut data = String::new();
+                match stdout_reader.read_line(&mut data) {
```
CatGarab: Will this block indefinitely until a line's worth of data is available? (and the same for stderr)
If so, could you use something like [this crate](https://docs.rs/timeout-readwrite/0.1.0/timeout_readwrite/) to add a timeout?

libs/shell-protocol/src/messages/stderr.rs
plauche: The empty message indicates that the process's stderr/stdout pipe has been closed.

libs/shell-protocol/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo hello
+echo world
```
CatGarab: nit: test file which probably doesn't need to be included

PR #196, opened 2018-10-11
Updating example mission apps

CatGarab:
- Updating the python example mission app to a hardware-independent query and mutation
- Updating the rust example to match the python example


CatGarab:
[Update example mission apps](https://trello.com/c/JMtdjFIA/119-update-example-mission-apps)

PR #202, opened 2018-10-18
Shell protocol stdin handling

plauche:


plauche:
[Shell protocol: write to process](https://trello.com/c/Xzgjzicv/111-shell-protocol-write-to-process)

libs/shell-protocol/src/messages/mod.rs
```diff
@@ -68,6 +73,8 @@ pub fn parse_message(message: ChannelMessage) -> Result<Message, ProtocolError>
         Ok(spawn::from_cbor(&message)?)
     } else if message.name == "stderr" {
         Ok(stderr::from_cbor(&message)?)
+    } else if message.name == "stdin" {
```
CatGarab: Not sure why I didn't notice before. Why did you choose to go with an `if-else` chain rather than `match`?

libs/shell-protocol/src/messages/stdin.rs
```diff
@@ -0,0 +1,76 @@
...
+use super::*;
+use channel_protocol::ChannelMessage;
+use error::ProtocolError;
+use serde_cbor::ser;
+
+/// CBOR -> Message::Stdin
+pub fn from_cbor(message: &ChannelMessage) -> Result<Message, ProtocolError> {
+    let data = match message.payload.get(0) {
+        Some(Value::String(data)) => data,
+        _ => {
```
CatGarab: According to the old lua docs, the stdin message is allowed to not specify a data value, similar to stdout and stderr

libs/shell-protocol/src/messages/mod.rs
plauche: I hadn't dug into match enough to see it working with strings, but it definitely cleans it up!

libs/shell-protocol/src/protocol.rs
```diff
@@ -136,6 +136,15 @@ impl Protocol {
                         .send(messages::pid::to_cbor(self.channel_id, process.id()?)?)?;
                 }
             }
+            messages::Message::Stdin { channel_id, data } => {
+                info!("-< {{ {}, stdin, {:?} }}", channel_id, data);
```
CatGarab: nit: typo `-<` -> `<-` :laughing:

PR #203, opened 2018-10-19
Adding bad entry error handling + fixing boot time registry creation

CatGarab:
This PR does two things:
1. If a registry app entry becomes corrupted for some reason, the entry will now be removed and its directory will be deleted
2. Creating the `active` symlink directory when the registry is created, so the `run_onBoot` logic will run correctly for pre-existing applications


CatGarab:
[App Service: Add error cleanup](https://trello.com/c/aeDiX3cA/128-app-service-add-error-cleanup)

PR #204, opened 2018-10-22
Updating app service mutation schema to match other services

CatGarab:
All other Kubos services use the format `{ success, errors, data }` for their mutation responses.

This PR updates the app service to have a similar format.


CatGarab:
[Update the applications service to use the standard `{success, error, data}` return field format for all mutations](https://trello.com/c/N6PvmIyg/18-update-the-applications-service-to-use-the-standard-success-error-data-return-field-format-for-all-mutations)

PR #205, opened 2018-10-23
Adding the ability to upgrade applications

CatGarab:
Currently it is possible to upgrade applications with the applications service by simply re-registering them. However, this only works if the application name remain the same.

The original intention was that the application name could be changed between versions and the application UUID would be used as the main identifier.

This PR updates the `register` mutation to allow a UUID to be passed in order to upgrade an existing application. The documentation has also been updated to reflect the changes.


CatGarab:
[App Service: Update `register` mutation to support app upgrades](https://trello.com/c/BxKOJHOL/130-app-service-update-register-mutation-to-support-app-upgrades)

PR #211, opened 2018-10-30
Expanding timestamp size to be a Double. Change units to seconds

CatGarab:
Milliseconds since unix epoch overflowed the current i32 storage size. GraphQL (and Juniper) doesn't support i64, meaning the next biggest available data size is f64.

As a result, the telem db timestamps have been updated to use fractional seconds


CatGarab:
[Fix telem DB timestamp](https://trello.com/c/DkpxLYMY/466-fix-telem-db-timestamp)

PR #212, opened 2018-11-01
Adding bad exit code to Python app example and tutorial

CatGarab:
Adding non-zero exit codes to the example Python mission app and the corresponding tutorial. This will become more important once the app service is capable of checking a started app's return code


CatGarab:
[Update Python AppAPI to return exit code](https://trello.com/c/3rogkmSe/132-update-python-appapi-to-return-exit-code)

PR #218, opened 2018-11-26
File transfer false positive bug

plauche:
- Fixes file client reporting positive if upload files
- Fixes bug in file service upload_single* tests

### DIFF
--- a/libs/file-protocol/src/protocol.rs
+++ b/libs/file-protocol/src/protocol.rs
@@ -538,7 +538,12 @@ impl Protocol {
                     }
                     State::Holding { count, prev_state } => {
                         if count > self.config.hold_count {
-                            return Ok(());
+                            match prev_state.as_ref() {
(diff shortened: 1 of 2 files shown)


plauche:
[File transfer reporting false positive](https://trello.com/c/cnvQ8LRn/122-file-transfer-reporting-false-positive)

libs/file-protocol/src/protocol.rs
```diff
@@ -538,7 +538,12 @@ impl Protocol {
                     }
                     State::Holding { count, prev_state } => {
                         if count > self.config.hold_count {
-                            return Ok(());
+                            match prev_state.as_ref() {
+                                State::Holding { .. } => return Ok(()),
+                                _other => {
+                                    return Err(ProtocolError::ReceiveTimeout);
```
plauche: @CatGarab do you think I should return `ReceiveTimeout` all the time or get more specific in this error?

PR #221, opened 2018-11-29
Converting shell service to use syslog for logging

CatGarab:
This is my plan for how all the services will be updated to use syslog for logging.

Worth noting, this does remove the terminal logger component. This behavior can be re-added by the user by setting up a logging policy which routes the messages to the console (or some similar policy):
`daemon.*    /dev/console`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -260,6 +260,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "error-chain"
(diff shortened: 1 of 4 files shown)


CatGarab:
[update services to use logging crate](https://trello.com/c/iKKSxhne/152-update-services-to-use-logging-crate)

PR #227, opened 2018-12-04
Adding logging setup to monitor service

CatGarab:
Also updating the `kubos-service` crate to use the base `log` crate's macros


CatGarab:
[update services to use logging crate](https://trello.com/c/iKKSxhne/152-update-services-to-use-logging-crate)

PR #232, opened 2018-12-05
Updating examples to include logging

CatGarab:


CatGarab:
[update mission application examples to use logging library/crate](https://trello.com/c/TJemW4xm/153-update-mission-application-examples-to-use-logging-library-crate)

examples/python-mission-framework/python-mission-app.py
```diff
@@ -1,27 +1,44 @@
...
-def on_command():
+def on_command(logger):
     
-    print "OnCommand logic"
+    logger.info("OnCommand logic")
 
 def main():
+
+    logger = logging.getLogger('mission-framework')
+    logger.setLevel(logging.INFO)
```
jacoffey3: Is this always ".INFO?" Does that control what gets sent to the file or what prints to stdout?

CatGarab: The level can be [any of these](https://docs.python.org/3/library/logging.html#logging-levels)

It controls what log messages are actually generated. So if you've got it set to info, but then do `logging.debug("debug message")`, that message gets totally ignored by the program

jacoffey3: So...this should definitely say debug then, because we have it in place to have debug level and we don't want the users to ignore that

PR #233, opened 2018-12-05
Updating GraphQL responses to match spec

CatGarab:
Updating our GraphQL responses to [match the spec](https://facebook.github.io/graphql/June2018/#sec-Response-Format)

- "msg" -> "data"
- "errs" -> "errors"
- Updated the case where an error is hit before any data is generated to return the errors array in an object, rather than just an array


CatGarab:
[Kubos Service GraphQL Spec Compliance](https://trello.com/c/sYwsywkY/170-kubos-service-graphql-spec-compliance)

PR #234, opened 2018-12-06
Allowing multiple files to be present in app registry dir

CatGarab:
Currently, the app service expects exactly two files to be present in the directory used to register an application: the `manifest.toml` file and the executable.

For Python applications, there is no executable, so it's entirely likely that users will want to register an application with more than one file.

This PR changes the registration logic to copy the entire contents of the directory specified when registering an app


CatGarab:
[Allow more than 2 files to be in a mission app directory](https://trello.com/c/TVp649NG/177-allow-more-than-2-files-to-be-in-a-mission-app-directory)

PR #239, opened 2018-12-07
Updating graphql response in various places

plauche:


apis/app-api/python/app_api.py
```diff
@@ -78,13 +78,13 @@ def _format(self, response, service):
             raise
 
         # Check that it follows GraphQL format
-        if 'msg' not in response or 'errs' not in response:
+        if 'data' not in response or 'errors' not in response:
```
CatGarab:
```suggestion
        if 'errors' not in response:
```

PR #240, opened 2018-12-07
[WIP] Initial Comms Service Implementation

srjustice:
This is a first attempt at creating a generic `comms-service` for KubOS. To see the requirements, assumptions, and constraints that influenced the `comms-service` contained in this PR, please refer to this [card](https://trello.com/c/k6vYVRfZ) on the KubOS Community Trello board.

The two main additions in this PR are the `comms-service` and the `ethernet-service`.

The `comms-service` is a generic library that allows any hardware service to be transformed into a gateway capable of sending/receiving data to/from the ground.

The `ethernet-service` is a sort of "proof of concept" for the `comms-service`. The `ethernet-service` uses the `comms-service` to enable users to use an Ethernet connection like a radio in order to communicate with a satellite running KubOS.


jacoffey3:
[Communication Service Spike](https://trello.com/c/k6vYVRfZ/52-communication-service-spike)

srjustice:
@jacoffey3 I already linked the card in the description!

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,315 @@
...
+const MAX_SIZE: usize = 65499;
+
+/// Type definition for a "read" function pointer.
+pub type ReadFn<T> = Fn(T) -> CommsResult<Vec<u8>> + Send + Sync + 'static;
+/// Type definition for a "write" function pointer.
+pub type WriteFn<T> = Fn(T, &[u8]) -> CommsResult<()> + Send + Sync + 'static;
+
+/// Struct that holds configuration data to allow users to set up a Communication Service.
+#[derive(Clone)]
+pub struct CommsControlBlock<T: Clone> {
```
plauche: I wonder if we should put any additional trait constraints on `T`. Requiring just `Clone` seems a bit too general. 

Is `T` effectively defining the socket type used to talk with the radio and any processes on the system? 

How would this work if we moved to a system where GraphQL services speak HTTP/TCP but other services still speak UDP?

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,315 @@
...
+            log_error(&data, CommsServiceError::InvalidChecksum.to_string()).unwrap();
+            continue;
+        }
+
+        // Update number of packets up.
+        log_telemetry(&data, TelemType::Up).unwrap();
+
+        // Spawn new message handler.
+        let conn_ref = comms.write_conn.clone();
+        let write_ref = comms.write[0].clone();
```
plauche: Is the assumption here that all of the `write` functions will be the same OR that `write[0]` will always be the appropriate one to use here?

services/comms-service/src/telemetry.rs
```diff
@@ -0,0 +1,46 @@
+use errors::*;
+use std::sync::{Arc, Mutex};
+
+/// Generic telemetry collected by the communication service.
+#[derive(Default, GraphQLObject)]
+pub struct CommsTelemetry {
```
plauche: I know we are already capturing errors but would it be good to also capture the number of `packets_up` and `packets_down` that had errors?

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,315 @@
...
+/// Type definition for a "write" function pointer.
+pub type WriteFn<T> = Fn(T, &[u8]) -> CommsResult<()> + Send + Sync + 'static;
+
+/// Struct that holds configuration data to allow users to set up a Communication Service.
+#[derive(Clone)]
+pub struct CommsControlBlock<T: Clone> {
+    /// Function pointer to a function that defines how to read from a gateway.
+    pub read: Option<Arc<ReadFn<T>>>,
+    /// Optional function pointers to functions that define methods for writing data over a gateway.
+    pub write: Vec<Arc<WriteFn<T>>>,
```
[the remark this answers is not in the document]
greer184: The reason that they are separate is that in our initial design, having downlink ports was optional, but if we want all services running the communication service to have the specified endpoints, this change would make sense.

services/comms-service/src/service.rs
greer184: The assumption is that the first write function will be the one used in the ground-to-satellite full loop.

Cargo.toml
```diff
@@ -30,6 +30,7 @@ members = [
 "services/app-service",
 "services/monitor-service",
 "services/comms-service",
```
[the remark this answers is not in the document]
srjustice: Ya, that makes sense. The only reason we put it in the services folder was because the `kubos-service`, which is also a backing library, is in the services folder.

services/comms-service/src/telemetry.rs
greer184: Those would be good to add to future telemetry items.

services/comms-service/src/service.rs
greer184: T is simply the connection interface you are writing across and receiving data upon. For the ethernet this is a UDP socket, but for our radios, they use UART connections. 

We could force users to implement a trait on those objects that includes functions that perform framing and fragmentation, but since they're defining the functions anyways, I'm not too sure whether or not having a tighter bound on T would be that useful.

In the case where GraphQL services speaks a different protocol other UDP, we would have to tweak the actual generic library to use those protocols as it simply assumes UDP the whole way. But this tweak wouldn't affect the read and write functions or T since those deal with constructing / deconstructing & farming / deframing of UDP packets. The generic comms-service itself deals with communicating with the rest of the services.

services/comms-service/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Each port in the list will be used by one downlink endpoint.
+    pub downlink_ports: Option<Vec<u16>>,
+    /// Timeout for the completion of GraphQL operations within message handlers (in milliseconds).
+    pub timeout: u64,
+    /// IP address of the computer that is running Agent Tom.
+    pub ground_ip: String,
+    /// Specifies the port to which Agent Tom is bound.
+    pub ground_port: Option<u16>,
+    /// Satellite's IP address.
+    pub satellite_ip: String,
```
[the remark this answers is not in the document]
greer184: Whichever the communication service that is running is bound to and the one that Major Tom is talking to from the ground. The IP addresses are solely used in the UDP checksums to verify the received packet is in the correct location.

srjustice:
@plauche Sure! Ya, that makes sense. We will add that in.

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,339 @@
...
+
+    // Take reader from control block.
+    let read = comms.read.unwrap();
+
+    loop {
+        // Read bytes from the radio.
+        let bytes = match (read)(comms.read_conn.clone()) {
+            Ok(bytes) => bytes,
+            Err(e) => {
+                log_error(&data, e.to_string()).unwrap();
```
plauche: It would be good to go through all of these bare `unwrap`s and decide whether or not we are ok with a `panic` happening there.

greer184: All unwraps that are present in the threads should be in the case of mutex poisoning.

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,339 @@
...
+pub struct CommsService;
+
+impl CommsService {
+    /// Starts an instance of the Communication Service and its associated background threads.
+    pub fn start<T: Clone + Send + 'static>(
+        control: CommsControlBlock<T>,
+        telem: Arc<Mutex<CommsTelemetry>>,
+    ) -> CommsResult<()> {
+        // Make sure a `read()` function and a `write()` function exist before starting the read thread.
+        if control.read.is_some() {
```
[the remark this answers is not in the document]
greer184: If you only want to downlink data from mission apps.

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,339 @@
...
+            if control.write.len() > 0 {
```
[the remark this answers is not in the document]
greer184: Yes.

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,339 @@
...
+        let sat_ref = comms.satellite_ip.clone();
+        let ground_ref = comms.ground_ip.clone();
+        let time_ref = comms.timeout.clone();
+        thread::spawn(move || {
+            handle_message(
+                data_ref, port, conn_ref, write_ref, packet, time_ref, sat_ref, ground_ref,
+            );
+        });
+
+        // Increment port number.
```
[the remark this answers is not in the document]
greer184: With the current logic, the packet would be dropped. If we checked binding outside the handler, then we could choose to drop the packet, we could wait for a port to open up, or we could buffer the packet for later when a port may be open.

services/comms-service/src/service.rs
```diff
@@ -0,0 +1,339 @@
...
+/// Type definition for a "read" function pointer.
+pub type ReadFn<T> = Fn(T) -> CommsResult<Vec<u8>> + Send + Sync + 'static;
+/// Type definition for a "write" function pointer.
+pub type WriteFn<T> = Fn(T, &[u8]) -> CommsResult<()> + Send + Sync + 'static;
+
+/// Struct that holds configuration data to allow users to set up a Communication Service.
+#[derive(Clone)]
+pub struct CommsControlBlock<T: Clone> {
+    /// Function pointer to a function that defines how to read from a gateway.
+    pub read: Option<Arc<ReadFn<T>>>,
```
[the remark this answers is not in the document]
greer184: For the radio we're using, we have multiple ways of downlink communication to the ground (store and forward, realtime) to accommodate different volumes of data coming from the satellite. Since we don't anticipate regularly sending data up to the satellite, we left it with a single uplink endpoint. In a lot of cases, you'll only have one uplink endpoint and one downlink endpoint, but we left the flexibility with the downlink endpoints since more scenarios involve using the downlink endpoints.

plauche:
This PR has been deprecated by #251. Same work but different PR :smile:

PR #246, opened 2018-12-20
Updating the app APIs to add logging initialization

CatGarab:


CatGarab:
[Update app APIs to use logging](https://trello.com/c/HpQBX24p/179-update-app-apis-to-use-logging)

PR #247, opened 2018-12-21
Running cargo clippy.

plauche:
Worked through a bunch of the feedback from `clippy`

For now we are intentionally ignoring these items due to needing more functional changes:
- [`clippy::type_complexity`](https://rust-lang.github.io/rust-clippy/master/index.html#type_complexity)
- [`clippy::new_ret_no_self`](https://rust-lang.github.io/rust-clippy/master/index.html#new_ret_no_self)
- [`clippy::implicit_hasher`](https://rust-lang.github.io/rust-clippy/master/index.html#implicit_hasher)
- `proc_macro_derive_resolution_fallback` - We are getting warnings from macros from juniper crate. We may just need to upgrade to resolve them.

### DIFF
--- a/libs/file-protocol/src/storage.rs
+++ b/libs/file-protocol/src/storage.rs
@@ -347,7 +347,7 @@ pub fn initialize_file(
     }
 
     store_meta(prefix, &hash, index)?;
-    fs::remove_file(&temp_path).is_err();
+    fs::remove_file(&temp_path);
 
     if let Ok(meta) = fs::metadata(source_path) {
         Ok((hash, index, meta.mode()))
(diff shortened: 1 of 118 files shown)


libs/file-protocol/src/storage.rs
```diff
@@ -349,7 +347,7 @@ pub fn initialize_file(
     }
 
     store_meta(prefix, &hash, index)?;
-    fs::remove_file(&temp_path);
+    fs::remove_file(&temp_path).is_err();
```
CatGarab: Not sure what this change is for

PR #250, opened 2019-01-02
Adding documentation for communication service framework

CatGarab:


CatGarab:
[Communication Service Admin Docs](https://trello.com/c/4toizDdX/205-communication-service-admin-docs)

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+
+The first layer will be whatever communication protocol the device requires.
+For example, AX.25 is frequently used as the header protocol for radio communication.
+
+Inside of this will be a UDP packet, most likely containing a GraphQL request intended for one of
+the KubOS or payload services.
+
+TODO: Packet diagram?
+
+Ground Communication
```
jacoffey3: AFAIK, there is also a way to have a UDP packet pass straight through without being assigned a wait time (eg, if something is going to the shell or file service). I do not know if this is something you overlooked or if it's something that didn't make it in.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+~~~~~~~~~~~~~~~~~~~~
+
+The communications service maintains a constant read thread which listens for messages from the
+ground via the communications device.
+
+Once a message is received, a message handler thread is spawned and assigned one of the available
```
jacoffey3: The destination port should be checked here and referenced with the configuration toml if it's a pass through or a transaction.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+
+.. note::
+
+    The communications service framework is currently only available in Rust. As a result, services
+    wishing to utilize the framework should also be written in Rust.
+
+Architecture
+------------
+
+TODO: Overall architecture diagram
```
srjustice: For the overall architecture diagram are you meaning something like the diagram you sent above? If so, I think that would be super helpful. That is what we tried to capture with the `Comms_Service_Conceptual_Diagram.png` we linked to the Comms service card on the community Trello. However, that diagram is slightly out of date. That was made when we only had a primary and secondary downlik endpoint instead of an arbitrary number of downlink endpoints.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+
+All packets sent to/from the communication device will be encapsulated in several layers.
+
+The first layer will be whatever communication protocol the device requires.
+For example, AX.25 is frequently used as the header protocol for radio communication.
+
+Inside of this will be a UDP packet, most likely containing a GraphQL request intended for one of
+the KubOS or payload services.
+
+TODO: Packet diagram?
```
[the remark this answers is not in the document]
srjustice: I definitely think a packet diagram would be helpful. Doesn't have to be super specific, but something that shows the encapsulating protocol-UDP packet-GraphQL query/mutation layers would be nice to see.

docs/services/comms-framework.rst
srjustice: @jacoffey3 This is definitely how we want it to work! However, we didn't get this functionality into our PR. If you guys added it, then it should definitely be mentioned.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+
+Ground Communication
+~~~~~~~~~~~~~~~~~~~~
+
+The communications service maintains a constant read thread which listens for messages from the
```
srjustice: You state that the read thread listens for messages from the ground via the _communications device_. I noticed that in our diagram we titled that lifeline as _Radio_. I didn't know if you might want to rename that lifeline to _Communications Device_ in order to illustrate that the comms service enables communication over more than just a radio.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+ground via the communications device.
+
+Once a message is received, a message handler thread is spawned and assigned one of the available
+handler UDP ports.
+This message handler examines the message to determine the internal message destination and then
```
srjustice: You may want to mention that the "internal message destination" comes from the destination port embedded in the UDP packet. If you don't think it is necessary to be that explicit, that is totally fine!

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+An example would be when sending a periodic health-and-status beacon.
+
+In this case, the communications service may be defined with a downlink endpoint thread (or multiple
+threads if more than one method may be used for downlink communication).
+
+Each endpoint is assigned its own UDP port and maintains a constant read thread which listens for
+messages from within the satellite which should be transmitted.
+
+When the endpoint's read thread receives a message, it wraps it up in a UDP packet and then sends
+it to the endpoint device, via the user-defined write function.
```
srjustice: I've noticed that the device the comms service manages is referred to by several different names throughout the document: "endpoint device", "device", "communications device", "hardware device". It might be best to pick one and use that everywhere.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+Configuration
+-------------
+
+Most missions tend to have a single communications device which communicates via a single uplink
+and single downlink method.
+However, some missions have more complex communications systems; for example, having a slower, but
+more reliable downlink method for transmitting a small health-and-status beacon and then a faster,
+but more error-prone method for transmitting more detailed telemetry information.
+
+The communications framework provides mechanisms to handle these more complex configurations.
```
srjustice: I've noticed the comms service is sometimes referred to as the "communications service", "communications service framework", "communications framework", and just "service". It might be best to pick one and use that everywhere.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+
+The service's `config.toml` file should contain the following parameters:
+
+- ``handler_port_min`` - Starting port used to define a range of ports that are used in the message
+  handlers that handle messages received from the ground
+- ``handler_port_max`` - Ending port used to define a range of ports that are used in the message
+  handlers that handle messages received from the ground
+- ``downlink_ports`` - (Optional) List of ports used by downlink endpoints that send messages to the
+  ground each port in the list will be used by one downlink endpoint
```
srjustice: I think there should be a period after "ground", and the "e" in "each" should be capitalized.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+- ``ground_ip`` - IP address of the ground gateway
+- ``ground_port`` - UDP port of the ground gateway
+- ``satellite_ip`` - IP address of the communications service
+
+The service which implements the framework should create a |CommsControlBlock|, which
+provides the final configuration to the main communication logic.
+It contains the following members:
+
+- ``read_conn`` - This is the connection structure which is used for reading from the communications
+  device, for example a UART port or UDP socket. It should implement the Read trait.
```
srjustice: Why should this implement the Read trait? The 'read()' method should not be called on the 'read_conn' itself. The 'read_conn' will be passed into the `read` function pointer and used within that function. At least, that is how we envisioned it working.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+- ``write_conn`` - This is the connection structure which is used for reading from the
+  communications device, for example a UART port or UDP socket. It should implement the Write trait.
```
srjustice: Same with this. Why should `write_conn` implement the Write trait? We envisioned the `write_conn` being passed into the `write` function and used within that function.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+device.
+If multiple devices are present in the system, a service should be created for each of them.
+
+The service should contain the following components:
+
+- A function capable of reading messages from the communications device
+  (note: this may be omitted in the case where the communications device will only be used for
+  transmitting data)
+- At least one function capable of writing messages to the communications device
+- Initialized connection structure/s capable of reading and/or writing from/to the communications
```
srjustice: Once again, we didn't envision the connection structures doing the actual reading/writing. We envisioned them being used to read/write from/to the communications device.

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+  device (this structure will be used by the previous read and/or write functions)
+- A |CommsControlBlock| structure containing all of the communications configuration
+  information
+- A |CommsTelemetry| structure which will be used to gather communication statisticss
```
srjustice: Typo in "statistics".

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+  device, for example a UART port or UDP socket. It should implement the Read trait.
+- ``write_conn`` - This is the connection structure which is used for reading from the
+  communications device, for example a UART port or UDP socket. It should implement the Write trait.
+- ``read`` - (Optional) A pointer to the function which should be used to read a message from the
+  communications device
+- ``write`` - A list of function pointers for all available ways that messages may be written to
+  the communications device
+- ``handler_port_min`` - Should be copied from the corresponding `config.toml` value
+- ``handler_port_max`` - Should be copied from the corresponding `config.toml` value
+- ``downlink_ports`` - Should be copied from the corresponding `config.toml` value or ``None``
```
srjustice: You may want to mention somewhere that if `downlink_ports` is specified, it must be the same length as `write`.

docs/services/comms-framework.rst
srjustice: Yup!

docs/services/comms-framework.rst
CatGarab: Read trait might be the wrong thing, but the code will look something like this:
```
fn read(conn: read_conn) -> Result<vec<u8>> {
    let msg = conn.read()?;
    return msg;
```

so, `read_conn` needs to have a read function available. i can't think of any connection example that *wouldn't* need it.

docs/services/comms-framework.rst
CatGarab: Nope, we have not added it. @jacoffey3 you'll want to create a card for future work

docs/services/comms-framework.rst
CatGarab: Leaving this alone for now, since that's behavior which hasn't been implemented yet

docs/services/comms-framework.rst
CatGarab: yeah, it'll probably be some hybrid of the two

docs/services/comms-framework.rst
srjustice: Ya for sure! You are totally right! `read_conn` needs to have some sort of function related to reading. The `read_conn` in our Ethernet implementation is an `Arc<UdpSocket>` that uses `recv_from()` for its read functionality. It does not implement the `Read` trait. The `std::io::Read` trait requires that you have a literal `read()` function, and the only constraint we put on `read_conn` is that it implements the `Clone()` trait.

docs/services/comms-framework.rst
CatGarab: I commonized a bunch. There are a couple instances where it's different that i want to keep

docs/services/comms-framework.rst
CatGarab: :+1: 
I've settled on the following:
- "communications service framework" - the library that y'all created
- "communications service" - the final executable

docs/services/comms-framework.rst
CatGarab: oh interesting, didn't know that. i'll update the docs :)

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,155 @@
...
+ground via the communications device.
+
+Once a message is received, a message handler thread is spawned and assigned one of the available
+handler UDP ports.
+This message handler examines the message to determine the internal message destination and then
+forwards it on to the appropriate service.
+The handler then waits for a reply (within a specified timeout duration), wraps the response in a
+UDP packet, and then sends the packet to the device for transmission.
+
+.. figure:: ../images/comms_from_ground.png
```
jacoffey3: In this figure and the next, framing is pictured. There should either be a note that framing is coming in a future release or it should be removed from the diagrams

CatGarab: :+1: good catch

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,227 @@
...
+first receiving a request for data.
+An example would be when sending a periodic health-and-status beacon.
+
+In this case, the communications service may be defined with a downlink endpoint thread (or multiple
+threads if more than one method may be used for downlink communication).
+
+Each endpoint is assigned its own UDP port and maintains a constant read thread which listens for
+messages from within the satellite which should be transmitted.
+
+When the endpoint's read thread receives a message, it wraps it up in a UDP packet and then sends
```
plauche: Just to clarify, it deconstructs the received UDP packet, removes the payload and packages it into a new UDP packet, correct?

docs/services/comms-framework.rst
```diff
@@ -0,0 +1,227 @@
...
+The service which implements the communications service framework should utilize only a single
+communication device.
+If multiple devices are present in the system, a service should be created for each of them.
+
+The service should contain the following components:
+
+- A function capable of reading messages from the communications device
+  (note: this may be omitted in the case where the communications device will only be used for
+  transmitting data)
+- At least one function capable of writing messages to the communications device
```
plauche: Could you also have a communications service which only reads from a communications device and does not write?

docs/services/comms-framework.rst
CatGarab: yes, however the first two steps are just the result of `socket.recv()`

docs/services/comms-framework.rst
CatGarab: [not currently](https://github.com/kubos/kubos/blob/master/libs/comms-service/src/service.rs#L110)

PR #252, opened 2019-01-04
Updating docs to reflect new default logging behavior

CatGarab:
Updating docs to match changes in kubos/kubos-linux-build#150


CatGarab:
[Changes per log testing](https://trello.com/c/fNPj8aiS/156-changes-per-log-testing)

PR #253, opened 2019-01-07
Adding example comm service impl and client

CatGarab:


CatGarab:
[Example comm service implementation with a UART](https://trello.com/c/3wKR6FzZ/207-example-comm-service-implementation-with-a-uart)

PR #254, opened 2019-01-09
Moving communications settings into main toml file

CatGarab:
Currently the communication service library expects the comms settings to be established in a custom config.toml file

This PR updates the library to use the standard config.toml file provided by the `kubos-system` crate (apis/system-api)

Also updates the config reading logic to take advantage of `Deserialize` to convert the `toml::Value` into a `CommsConfig` structure, rather than needing to explicitly parse all the args


CatGarab:
[Comm service tweaks](https://trello.com/c/g2ESdw5H/214-comm-service-tweaks)

PR #256, opened 2019-01-12
Upgrade to Rust 2018 Edition

cantino:
Since I was [upgrading to Rust 2018 Edition](https://rust-lang-nursery.github.io/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html) on a personal project, I figured I'd do it here too. This PR is the result of running the following commands:

* `cargo kubos -c fix -- --edition --allow-dirty`
* `cargo kubos -c fix -- --edition-idioms --allow-dirty`
* `cargo kubos -c test`

The vast majority of the changes were adding `crate::` before module paths, as required in 2018 edition. The only other substantial change that I saw was `try!` becoming `r#try!` as a [raw identifier](https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/raw-identifiers.html).

I added `edition = "2018"` in each of the workspace `Cargo.toml` files individually because [it doesn't appear to be supported in the root](https://github.com/rust-lang/cargo/issues/5784).

Only test failure was `thread 'large_down' panicked at 'assertion failed: result.is_ok()', services/file-service/tests/download.rs:316:5`, which failed for me before any of these changes as well.

The only other change is adding a `.gitignore` file in `services/file-service/` to ignore the sub-directories `service` and `client` that appear to be created by tests. (These end up looking like `services/file-service/service/storage/22daf71d5cabad0a8c863326bafe27dc/12345`. Please confirm!)


cantino:
Whew, so that actually took a lot longer than I expected.

I've removed instances of `extern crate` with the exception of a few cases where I left `#[macro_use] extern create diesel` or `#[macro_use] extern create juniper` because they both have a bunch of private internal macros that don't play well with Rust 2018. See https://github.com/diesel-rs/diesel/pull/1787. There are also some warnings now that look like this:
```
warning: cannot find type `Entry` in this scope
  --> apis/telemetry-db-api/src/models.rs:20:17
   |
20 | #[derive(Debug, Queryable, Serialize, Deserialize)]
   |                 ^^^^^^^^^ names from parent modules are not accessible without an explicit import
   |
   = note: #[warn(proc_macro_derive_resolution_fallback)] on by default
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #50504 <https://github.com/rust-lang/rust/issues/50504>


warning: cannot find type `telemetry` in this scope
  --> apis/telemetry-db-api/src/models.rs:28:10
   |
28 | #[derive(Insertable)]
   |          ^^^^^^^^^^ names from parent modules are not accessible without an explicit import
   |
   = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
   = note: for more information, see issue #50504 <https://github.com/rust-lang/rust/issues/50504>


warning: cannot find type `subsystem` in this scope
 --> <::diesel::macros::__diesel_column macros>:5:33
  |
5 | derive ( Debug , Clone , Copy , QueryId , Default ) ] pub struct $ column_name
  |                                 ^^^^^^^ names from parent modules are not accessible without an explicit import
  |
  = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
  = note: for more information, see issue #50504 <https://github.com/rust-lang/rust/issues/50504>
```
I believe it's this issue: https://github.com/diesel-rs/diesel/issues/1785

I'm not sure if `warning: cannot find type 'telemetry' in this scope` means that the `#[table_name = "telemetry"]` line at `apis/telemetry-db-api/src/models.rs:28` isn't working, or if it's a confusing warning. Tests failures don't seem related, but I'm not sure.

I updated some doc comments. Do they get tested on CI?

On master, my only test failure is `test_delete_le`, but on this branch:

```
---- test_delete_le stdout ----
thread 'test_delete_le' panicked at 'assertion failed: `(left == right)`
  left: `Object({"data": Object({"delete": Object({"entriesDeleted": Null, "errors": String("attempt to write a readonly database"), "success": Bool(false)})}), "errors": String("")})`,
 right: `Object({"data": Object({"delete": Object({"entriesDeleted": Number(9), "errors": String(""), "success": Bool(true)})}), "errors": String("")})`', services/telemetry-service/tests/test_delete.rs:177:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.

---- test_delete_ge stdout ----
thread 'test_delete_ge' panicked at 'recv function failed: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', services/telemetry-service/tests/utils/mod.rs:144:19

---- test_delete_parameter stdout ----
thread 'test_delete_parameter' panicked at 'recv function failed: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', services/telemetry-service/tests/utils/mod.rs:144:19

---- test_delete_subsystem stdout ----
thread 'test_delete_subsystem' panicked at 'recv function failed: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', services/telemetry-service/tests/utils/mod.rs:144:19


failures:
    test_delete_ge
    test_delete_le
    test_delete_parameter
    test_delete_subsystem
```

They seem intermittent? Not sure how the changes here could have caused them.

CatGarab:
- Doc comments do get tested during CI
- I'm fine with leaving the `extern crate` statements for the troublesome dependencies
- I'll look into the telemetry stuff

PR #259, opened 2019-01-17
Comms client over UART

CatGarab:


CatGarab:
[Example comms client](https://trello.com/c/K4dMmeNm/232-example-comms-client)

PR #262, opened 2019-01-18
Tweaking unit test config dir

CatGarab:
The current `config.toml` temporary path occasionally goes out-of-scope before the service has a chance to read it, causing whatever test to fail


CatGarab:
[Telemetry service unit tests periodically failing CI](https://trello.com/c/MW6oAORP/215-telemetry-service-unit-tests-periodically-failing-ci)

PR #269, opened 2019-01-22
Updating Python App API Documentation

CatGarab:
- Fixing python app api doc generation. Sphinx can auto-generate docs based on the code documentation, but the app api needs to be in the system path. So, I updated the `prep-docs.sh` script to do the installation before making the `sphinx-build` call
- Adding some actual code documentation


CatGarab:
[Python App API doc](https://trello.com/c/5ErN87du/87-python-app-api-doc)

PR #270, opened 2019-01-22
Removing CSP support and cleaning up C HAL

CatGarab:
- Deleting `libcsp`
- Deleting `tinycbor` (it was only used by libcsp)
- Updating all example projects to remove the `csp` dependency (most didn't actually need it anyways)
- Updating `kubos-hal` to remove CSP dependency and unsupported code
  - Removing the SPI, UART, GPIO, and SDIO modules
  - The I2C HAL is the only module that's actually functional for Linux targets (everything else was just stubbed out back in the way back to maintain compatibility with the FreeRTOS targets)


CatGarab:
[Give CSP a final Death](https://trello.com/c/gETB9WYe/42-give-csp-a-final-death)

khldragon:
hello everyone, I am confused why  Give CSP a final Death? what is the alternative solution？

plauche:
Hi @khldragon! There are a few reasons why we decided to move past CSP:

- We no longer support FreeRTOS, just Linux, so we did not need to OS abstraction functionality that CSP provided.
- In moving to Linux we shifted our system architecture to a microservices model and now rely on standard networking protocols for our IPC, rather than using CSP for IPC between tasks in a monolithic process.
- We decided to embrace and prioritize Rust when developing our internal middleware services. This meant going through our various services and libraries and removing/replacing any C code bases which weren't essential.

What are you trying to accomplish which requires CSP?

PR #274, opened 2019-01-24
Migrating to Python3

CatGarab:
Migrating all the application-related code and our boiler-plate service code to Python3.5 (this is what's included with our current version of BuildRoot). I tested the new code against both Python3.5 and our current Python2.7 and it seems like the new code is backwards-compatible.

Most of the changes are just updating the `print` syntax.

Most notable is the way strings are declared/handled:
- Python 2.7 `str` -> 3.5 `bytes`
- 2.7 `unicode` -> 3.5 `str`
- `bytes` in 2.7 is just an alias to `str`
- `unicode` does not exist at all in 3.5

The mission app examples didn't need to be updated

Also, tweaking the kubos service framework to correctly set up logging

### DIFF
--- a/apis/app-api/python/app_api.py
+++ b/apis/app-api/python/app_api.py
@@ -50,15 +50,15 @@ class Services:
         if service not in self.config:
             raise KeyError(
                 "Service name invalid. Check config file for service names.")
-        if type(query) not in [str, unicode]:
-            raise TypeError("Query must be str or unicode.")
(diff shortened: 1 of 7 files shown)


CatGarab:
[Move Kubos Service Framework to Python 3](https://trello.com/c/DYeqTlBJ/51-move-kubos-service-framework-to-python-3)

jacoffey3:
@srjustice Aren't you excited?

PR #279, opened 2019-01-29
Converting Python services to use HTTP instead of UDP

CatGarab:


CatGarab:
[Update Python `kubos-service` library to use HTTP](https://trello.com/c/NJtwrd0G/250-update-python-kubos-service-library-to-use-http)

examples/python-service/config.toml
```diff
@@ -1,3 +1,3 @@
 [example-service.addr]
-ip = "0.0.0.0"
+ip = "192.168.33.10"
```
jacoffey3: Why this?

libs/kubos-service/kubos_service/config.py
```diff
@@ -7,7 +7,7 @@
 import argparse
 import toml
 
-DEFAULT_IP = "127.0.01"
+DEFAULT_IP = "127.0.0.1"
```
jacoffey3: wow...that went undetected....

examples/python-service/config.toml
CatGarab: Whoops, that shouldn't have gotten merged

libs/kubos-service/kubos_service/config.py
plauche: lol wow

PR #281, opened 2019-02-01
Added CI cross compiling checks

plauche:
This will add `cargo check` for the mbm2 and iobc targets to our CI.


plauche:
[Add CI checks for cross compiler](https://trello.com/c/4WLwaJCe/270-add-ci-checks-for-cross-compiler)

PR #288, opened 2019-02-04
Removing yotta + cruft from Docker

plauche:


plauche:
[Remove yotta from docker](https://trello.com/c/wPmPUN34/268-remove-yotta-from-docker)

tools/dist/Dockerfile
```diff
@@ -1,33 +1,25 @@
-# kubos/kubos-dev:0.1.8
+# kubos/kubos-dev:0.1.9
```
CatGarab: nit: This is slightly off because we forget to update it sometimes. Make it match what the new version will actually be (based on the [docker hub version list](https://cloud.docker.com/u/kubos/repository/docker/kubos/kubos-dev)

tools/dist/Dockerfile
```diff
@@ -1,33 +1,25 @@
...
-MAINTAINER kyle@kubos.co
-
-RUN add-apt-repository -y ppa:george-edison55/cmake-3.x
+MAINTAINER catherine@kubos.com, ryan@kubos.com
 
 RUN apt-get update -y
 
 RUN apt-get upgrade -y python2.7
-RUN apt-get install -y build-essential libssl-dev libffi-dev libhidapi-hidraw0 clang
-RUN apt-get install -y python-setuptools build-essential ninja-build python-dev libffi-dev libssl-dev
```
CatGarab: might need `libssl-dev` for the new HTTP services. not sure.

plauche: Let me know if we need to add this back in.

PR #280, opened 2019-01-31
Removing last traces of yotta code/tools

plauche:

PR #289, opened 2019-02-04
Removing rest of yotta code+tools

plauche:
This is equivalent to #280 but I forgot to point that PR at master so here we are :penguin:


plauche:
[Remove yotta from kubos repo](https://trello.com/c/9M2lPHcR/266-remove-yotta-from-kubos-repo)

PR #293, opened 2019-02-11
Updating the app APIs to query via HTTP

CatGarab:
Updated the Rust and Python app APIs to do graphql requests over http instead of udp.

Also in this PR:
- Added a new flag `PKG_CONFIG_ALLOW_CROSS` to circleci. The Rust app API updates brought in a crate which needs the flag in order to be able to cross-compile
- Updated circleci config to use Python3 for everything
- Fixing some python unit tests
- Updating our Dockerfile to include the new python dependency (`responses`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
         - checkout
         - run:
             name: Cargo check for BBB
-            command: cargo check --target arm-unknown-linux-gnueabihf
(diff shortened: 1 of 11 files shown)


CatGarab:
[Update app APIs to use HTTP for queries](https://trello.com/c/i00QkIRB/273-update-app-apis-to-use-http-for-queries)

PR #297, opened 2019-02-15
Pinning rust version to 1.32.0

plauche:


plauche:
[Pin rust versions in Vagrant and Docker](https://trello.com/c/1h2gQ7cp/184-pin-rust-versions-in-vagrant-and-docker)

PR #302, opened 2019-02-19
Updating comms service config validation

CatGarab:
- Making `ground_ip` and `satellite_ip` required config parameters
- Enforcing certain config rules as soon as possible, rather than during the `start()` logic
  - Must have a `write()` function
  - If `downlink_ports` is present, `ground_port` must also be present
  - Number of downlink ports and number of write functions should be the same


libs/comms-service/src/tests/config.rs
```diff
@@ -0,0 +1,167 @@
+//
```
plauche: Dedicated test files should be in `comms-service/tests`. This will also eliminate the need for `tests/mod.rs` for including this test file.

libs/comms-service/src/service.rs
```diff
@@ -115,44 +159,36 @@ impl CommsService {
...
-                            &write_ref,
-                            ground_ip,
-                            sat_ip,
-                            ground_port,
-                        );
-                    });
-                }
-            } else {
-                return Err(CommsServiceError::ParameterLengthMismatch.into());
+            for (_, (port, write)) in ports.iter().zip(control.write.iter()).enumerate() {
```
plauche: How does this handle the situation where `control.write` and `ports` are different lengths? It looks like we aren't returning that specific error anymore.

CatGarab: I moved the check to `CommsControlBlock::new()` and changed the error to `CommsServiceError::ConfigError`

libs/comms-service/src/tests/config.rs
CatGarab: These are unit tests, rather than integration tests, so can belong in this folder

CatGarab:
[The ethernet-service throws an incorrect MissingWriteMethod error.](https://trello.com/c/D327qfrc/263-the-ethernet-service-throws-an-incorrect-missingwritemethod-error)

PR #304, opened 2019-02-26
Adding try/wait logic to app service

CatGarab:
Updating the app service to wait a moment after kicking off an application to see if it immediately fails or not.


CatGarab:
[App service: Do `try_wait` after calling application to check if the app immediately failed](https://trello.com/c/elcXuijU/70-app-service-do-trywait-after-calling-application-to-check-if-the-app-immediately-failed)

PR #305, opened 2019-02-27
Removing UUID from the app service

CatGarab:


CatGarab:
[Get rid of uuid in the mission app service and just use name to reference](https://trello.com/c/XduHe5MA/245-get-rid-of-uuid-in-the-mission-app-service-and-just-use-name-to-reference)

services/app-service/src/app_entry.rs
```diff
@@ -34,14 +38,16 @@ pub struct AppMetadata {
 /// Kubos App struct
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct App {
-    /// The generated UUID for the application
-    pub uuid: String,
+    /// The name of the application
+    pub name: String,
     /// The process ID of the application, if it's currently running (0 otherwise)
     pub pid: u32,
```
plauche: Does it make sense to keep `pid` in the `App` struct? Because we don't tract the life cycle of the running application, we can't set the `pid` back to `0`  when it stops running, so that value can't really be trusted.

I know this isn't in the scope of this PR but I wanted to raise the question.

CatGarab: :+1: I'll take it out. Nothing uses it.

docs/tutorials/app-register.rst
```diff
@@ -160,9 +152,8 @@ The response should like this::
           "errors": "",
           "entry": {
             "app": {
-              "uuid": "5eb20cf9-3b18-4713-b03f-681f1c1ca4b5",
-              "name": "my-mission-app.py",
-              "path": "/home/system/kubos/apps/5eb20cf9-3b18-4713-b03f-681f1c1ca4b5/1.0/my-mission-app.py"
+              "name": "my-mission-app",
+              "path": "/home/system/kubos/apps/my-mission-app/1.0/my-mission-app.py"
```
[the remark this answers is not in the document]
CatGarab: :+1: Renaming to `executable`

services/app-service/src/app_entry.rs
CatGarab: FYI: I left the `pid` response field in the `startApp` mutation

PR #306, opened 2019-02-27
Removing app dir after uninstalling last version

CatGarab:


CatGarab:
[App Service: Uninstalling an app should remove the app folder](https://trello.com/c/NqJCN9A6/176-app-service-uninstalling-an-app-should-remove-the-app-folder)

services/app-service/src/registry.rs
```diff
@@ -327,6 +327,11 @@ impl AppRegistry {
             errors = Some(format!("Failed to remove app directory: {}", error));
         }
 
+        // If that was the last version, also remove the parent directory.
+        // (If this call fails, it's probably because the directory wasn't empty because some
+        // version of the app still exists, so ignore the error)
+        let _ = fs::remove_dir(format!("{}/{}", self.apps_dir, app_name));
```
plauche: Do we know for sure that this was the last version? If we do then can we run `remove_dir_all` to ensure all files are cleaned up?

CatGarab: In theory, there are only two possible cases:
1. We just installed the last version of the app. In that case, we just did `remove_dir_all` for the version directory, so the parent app_name directory is now empty and `remove_dir` finishes successfully.
2. There are still other versions of the app present, so the parent app_name directory is not empty, so `remove_dir` fails.

I can add logic to read the app registry to see if we're uninstalling the last version, but I don't think there's much of a benefit

PR #308, opened 2019-02-28
Improving recovery if app install fails

CatGarab:
- Moving the logic to mark the old version of an app as inactive until the end of the installation process
- Adding logic to clean up the new app directories if the install fails

**note** I couldn't figure out how to test the directory cleanup. Installs should only really fail at that point if some weird system error pops up.


CatGarab:
[App Service: If an install fails, previous version should be still active.](https://trello.com/c/qm5Qr72P/32-app-service-if-an-install-fails-previous-version-should-be-still-active)

CatGarab:
[If an install fails, any files already transferred should be cleaned up](https://trello.com/c/OfH56LYC/287-if-an-install-fails-any-files-already-transferred-should-be-cleaned-up)

PR #313, opened 2019-03-13
Correcting file transfer client docs

CatGarab:
Fixing the syntax and examples in the file transfer client tutorial.

Also fixing the shell client tutorial syntax, but the actual examples for that one are correct


CatGarab:
[Here is from the tutorial on kubos-file-client:](https://trello.com/c/4S18KNtm/514-here-is-from-the-tutorial-on-kubos-file-client)

PR #316, opened 2019-03-18
Adding comms service integration tests

plauche:
Target test cases
- Downlink to ground
- Ground to service, without response
- Ground to service, with response via downlink
- Ground to service, with response via handler


plauche:
[Comms service integration tests](https://trello.com/c/UY55dxim/295-comms-service-integration-tests)

PR #322, opened 2019-03-28
Clydespace EPS 3G Post-Test Updates

CatGarab:
Fixing a few things now that I can actually test the EPS service on hardware.

Main changes:
- Updating the I2C HAL to do variable-length reads
- Adding an inter-command delay
- Adding big-endian -> native endian (little endian for MBM2) conversion for telemetry readings


CatGarab:
[EPS Testing](https://trello.com/c/iczryZ49/529-eps-testing)

PR #318, opened 2019-03-26
Modify comms-service to allow HTTP transactions

plauche:
- Adds integration tests around uplink and downlink behaviour
- Modifies uplink behaviour to allow communication with HTTP-based services
- Removes `handler_port_min/max` in favor of `max_num_handlers`

PR #326, opened 2019-04-04
Creating generic service testing utility

plauche:
In the process of #318 we wanted a test against one of our actual services. This was initially done in a way very specific to that service, but seems like a useful thing for other integration tests.

This PR pulls out that work and is the beginning of a method for writing integration tests which interact directly with other binary crates.


plauche:
[Crate testing method](https://trello.com/c/P3IOi2ps/791-crate-testing-method)

PR #338, opened 2019-04-30
Adding 'ping' query to all services

CatGarab:
A few of our services don't include the "ping" query. It's super nice to have, so I'm adding it.


CatGarab:
[OBC Housekeeping App](https://trello.com/c/QtdLdG3p/834-obc-housekeeping-app)

PR #340, opened 2019-05-02
Misc Tweaks

CatGarab:
Fixing a few issues which I found.
Adding the `globalstar_status` field to the supMCU service, which gives the current status of the NSL simplex

Note: CI ran into the "out of memory" issue we've had before, so I reduced the concurrent processes argument from 8 to 7


docs/services/telemetry-db.rst
```diff
@@ -117,23 +117,24 @@ The ``insert`` mutation can be used to add an entry to the telemetry database.
...
 ~~~~~~~~~~~
 
-The generated timestamp value will be the current system time in milliseconds.
+The generated timestamp value will be the current system time in seconds.
 The database uses the combination of ``timestamp``, ``subsystem``, and ``parameter`` as the primary key.
 This primary key must be unique for each entry.
 
-    - As a result, any one subsystem parameter may not be logged more than once per millisecond.
+    - As a result, there may be issues when attempting to log any one subsystem parameter more than
+      once per millisecond
```
plauche: Should this be `more than once per second` if the timestamp only has seconds resolution?

CatGarab: It's a float, so the units are seconds, but technically fractional seconds? Not entirely sure if that's a real phrase...

PR #347, opened 2019-05-15
Improving app service edge cases

CatGarab:
1. If the last version of an app is uninstalled, its active symlink will now be removed
2. If, for some reason, the active symlink directory stops existing, it will be recreated during the next app registration


CatGarab:
[If the "active" directory doesn't exist when you register a mission app, the registration fails](https://trello.com/c/cpdWDrLW/753-if-the-active-directory-doesnt-exist-when-you-register-a-mission-app-the-registration-fails)

CatGarab:
[Apps service uninstall can create orphan active symlink](https://trello.com/c/EZWrcjGQ/790-apps-service-uninstall-can-create-orphan-active-symlink)

PR #349, opened 2019-05-16
Correcting telem db table creation

CatGarab:
The telemetry database uses fractional seconds since UNIX epoch for its timestamps. Basically everywhere, this value is stored as a double.

This PR fixes the one remaining place that it's still an integer value.


CatGarab:
[Investigate telem db schema](https://trello.com/c/uw0P8f2V/763-investigate-telem-db-schema)

PR #351, opened 2019-05-16
Adding testing and docs for app args

CatGarab:
We've had partial support for Kubos app args for a while.
This PR finishes that functionality and adds documentation


CatGarab:
[Fix Rust app API passthrough args](https://trello.com/c/bRPuRQBg/765-fix-rust-app-api-passthrough-args)

.circleci/config.yml
```diff
@@ -44,6 +44,7 @@ jobs:
         - image: kubos/kubos-dev:latest
       steps:
         - checkout
+        - run: cd apis/app-api/python && pip3 install .
```
plauche: Should we just install the `app-api` in the `Dockerfile` instead?

CatGarab: I don't think so. If the app API changes, this will immediately get (and verify) the updates. The Dockerfile will be stuck with whatever version it was last built with

PR #357, opened 2019-05-21
Updating the app service to change to app's dir before kicking it off

CatGarab:
When registering an application, we currently copy the entire given directory. This allows users to register an app with multiple files. This is mostly important for Python, but does also give the user the ability to include auxiliary files (ex. a custom config file for something).

This PR updates the app service to change its working directory to the specified application's directory before kicking it off during a ``startApp`` request. This allows users to specify the files they included with their app using relative paths.

Note: Multi-file Python actually works without the app service changes, but I did expand the Python test project in order to verify this

### DIFF
--- a/services/app-service/src/registry.rs
+++ b/services/app-service/src/registry.rs
@@ -552,6 +552,12 @@ impl AppRegistry {
             return Err(AppError::StartError { err: msg });
         }
 
+        // Change our current directory to the app's directory so that it can access any
+        // auxiliary files with relative file paths
+        let cwd = ::std::env::current_dir()?;
+        let parent_dir = app_path.parent().unwrap_or(&cwd);
+        let _ = ::std::env::set_current_dir(parent_dir);
+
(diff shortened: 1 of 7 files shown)


CatGarab:
[Run directory isn't the app directory when started by the mission application service](https://trello.com/c/zO5coLHY/715-run-directory-isnt-the-app-directory-when-started-by-the-mission-application-service)

services/app-service/src/registry.rs
```diff
@@ -552,6 +552,12 @@ impl AppRegistry {
             return Err(AppError::StartError { err: msg });
         }
 
+        // Change our current directory to the app's directory so that it can access any
+        // auxiliary files with relative file paths
+        let cwd = ::std::env::current_dir()?;
+        let parent_dir = app_path.parent().unwrap_or(&cwd);
```
jacoffey3: Make a warning message if the cwd can't be found/retrieved

PR #358, opened 2019-05-21
Adding example OBC housekeeping app

CatGarab:


apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+This application is written in such a way that it can be run on any supported OBC with minimal
+changes.
+Please refer to the following 'Configuration' section for settings which should be updated for your
+particular system prior to execution.
+
+Configuration
+-------------
+
+All configuration options are defined as constant values at the top of the `src/main.rs` file.
```
cantino: Should this be configurable in a toml / yaml file instead of by modifying the source? Or am I misunderstanding?

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Before this application can be run on your particular stack, a few configuration options should be
+set:
+
+- ``USER_PARTITION`` - (Default: ``"/dev/mmcblk1p4"``) The device name of the user data partition.
+  The default value is valid for BBB and MBM2 target OBCs
+- ``MEM_TOTAL`` - (Default: ``515,340``) The amount of RAM (in bytes) present in the OBC.
+  The default value is valid for BBB and MBM2 target OBCs
+- ``COMMS_SERVICE`` - (Default: ``"ethernet-service"``) The name of the comms service which should
```
cantino: Novice question: if the disk is corrupted, won't the system fail to boot and end up running an earlier version of the OS or the recovery image?

CatGarab:
[Make OBC housekeeping app public](https://trello.com/c/v0rG6WnH/848-make-obc-housekeeping-app-public)

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
...
+The ``clean_db`` function deletes all telemetry database entries which have exceeded the threshold
+age (default: 1 week).
+
+Check Memory Usage
+~~~~~~~~~~~~~~~~~~
+
+The ``check_mem`` function gets the current percentage of RAM and disk (permanent storage) which are
+in use and checks them against the given threshold values (nominal, high, critical).
+
+If the RAM usage percentage exceeds the critical threshold (default: 80%), then the system is
```
cantino: (I could imagine wanting this check to run more often than hourly, but the other ones needing to be less frequent.)

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Check Filesystem Status
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``check_fs`` function verifies that the system can still read and write to/from the user data
+partition.
+If this check fails, then we can determine that the file system has become corrupted.
+In this case, an emergency beacon will be sent out over the specified communications device.
+
+Other failed checks simple write an error message to the system logs, however these logs normally
```
cantino: *simply*

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
...
+are no longer accessible, so we send a beacon instead.
+
+Ping Services
+~~~~~~~~~~~~~
+
+All Kubos services provide a "ping" request as part of their GraphQL schema.
+The ``ping_services`` function submits this ping request to all defined services and records which
+services fail to return the "pong" response.
+
+This function (located in `src/ping.rs`) should be updated to include all hardware services present
```
jacoffey3: It can't pull that from the `config.toml`?

cantino: Another thing I could imagine being in yaml / toml.

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
...
+set:
+
+- ``USER_PARTITION`` - (Default: ``"/dev/mmcblk1p4"``) The device name of the user data partition.
+  The default value is valid for BBB and MBM2 target OBCs
+- ``MEM_TOTAL`` - (Default: ``515,340``) The amount of RAM (in bytes) present in the OBC.
+  The default value is valid for BBB and MBM2 target OBCs
+- ``COMMS_SERVICE`` - (Default: ``"ethernet-service"``) The name of the comms service which should
+  be used to send an emergency beacon if the OBC's filesystem becomes unrecoverably corrupted.
+  This beacon may be disabled by making this value an empty string (``""``).
+- ``DOWNLINK_PORT`` - (Default: ``14011``) The specific comms service port which the emergency
```
jacoffey3: Can this be pulled from the `config.toml`?

apps/obc-hs/src/check_mem.rs
```diff
@@ -0,0 +1,105 @@
...
+use super::*;
+use std::process::Command;
+
+const OBC_TELEMETRY: &str = r#"{
+    memInfo {
+        available
+    }
+}"#;
+
+pub fn check_mem() -> Result<(), Error> {
```
jacoffey3: It should be documented that this requires the monitor service to be running

apps/obc-hs/src/check_fs.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    Ok(())
+}
+
+fn send_error() -> Result<(), Error> {
+    // Get the comms service IP address
+    let config = ServiceConfig::new(COMMS_SERVICE);
+    let host = config.hosturl();
+    let mut host_parts = host.split(':').map(|val| val.to_owned());
```
cantino: If this is a IP and port pair, should we use the given port instead of DOWNLINK_PORT?

apps/obc-hs/src/check_fs.rs
```diff
@@ -0,0 +1,81 @@
...
+    let comms_ip = host_parts.next().unwrap_or_else(|| "0.0.0.0".to_owned());
```
cantino: Should we log when defaulting to 0.0.0.0?

apps/obc-hs/README.rst
CatGarab: You're allowed to define multiple downlink ports within the comms service. There's no way to dynamically determine which one should be used here

apps/obc-hs/src/check_mem.rs
```diff
@@ -0,0 +1,105 @@
...
+        warn!("RAM usage high: {}%", ram_in_use);
+    } else {
+        error!("RAM usage too high: {}%. Triggering reboot", ram_in_use);
+        // Reboot the system not-nicely. If we're at this point, there's probably a rogue process
+        // that's hogging all the system resources and not playing nicely with others.
+        Command::new("reboot").arg("-f").status()?;
+    }
+
+    // Check disk space usage
+    // Get the % of the user data partition that's free
```
cantino: Out of curiosity, why do we use the monitor-service for memory but go directly to disk with `df` for disk?

apps/obc-hs/src/check_mem.rs
```diff
@@ -0,0 +1,105 @@
...
+            .filter_map(|&elem| {
+                if elem.is_ascii_digit() {
+                    Some(elem as char)
+                } else {
+                    None
+                }
+            })
+            .collect::<String>();
+
+        percent.parse::<u8>().unwrap_or(100)
```
cantino: If you can't parse, maybe log the output from `df`?

apps/obc-hs/src/check_mem.rs
```diff
@@ -0,0 +1,105 @@
...
+    // Check disk usage with respect to threshold
+    if disk_in_use < DISK_NOMINAL {
+        info!("Disk usage nominal: {}%", disk_in_use);
+    } else if disk_in_use < DISK_HIGH {
+        info!("Disk usage high, but acceptable: {}%", disk_in_use);
+    } else if disk_in_use < DISK_CRITICAL {
+        warn!("Disk usage high: {}%", disk_in_use);
+    } else {
+        error!("Disk usage too high: {}%. Triggering cleanup", disk_in_use);
+        // Delete everything from the database that's more than the critical age threshold
```
cantino: A restart would also probably free some space because `/tmp` would get cleaned? Maybe?

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,96 @@
+OBC Housekeeping App
+====================
+
+This application is responsible for checking the operating status of the main OBC and taking
```
cantino: Does anything monitor this application itself and restart it if it crashes?

jacoffey3: Not currently - we have an outstanding item to monitor applications, but we haven't had time to address it

apps/obc-hs/README.rst
CatGarab: The root filesystem and the user data are kept in two separate partitions. If the root fs partition becomes corrupted, yes, the OS recovery process will kick in an fix it. However, the OS recovery process specifically doesn't affect the user data partition

If the user data partition becomes corrupted, we have to do manual intervention. We run `fsck` at boot to help prevent this, but that can't handle all cases.

apps/obc-hs/src/check_mem.rs
CatGarab: Because the monitor service doesn't currently have the ability/functionality to check disk space (not that it couldn't, we just haven't added it)

apps/obc-hs/src/check_fs.rs
CatGarab: Fair question. The `hosturl` IP/port pair tells you the port that the service is listening on for GraphQL requests.

The comms service has additional listening ports which are used for receiving messages from other system processes, packing them up, and then sending them down to the ground (ie the downlink port/s).

To summarize, we're using a secondary port here, not the primary GraphQL port

apps/obc-hs/README.rst
CatGarab: I can kind of see that, too. I'm gonna leave it alone for now, but we can always change it in the future

apps/obc-hs/README.rst
CatGarab: ...probably :sweat_smile: but I'm not entirely sure.

For our services, yes, everything like this is controlled via the config.toml file. However, we intend for services to be able to be built and plopped into any system without needing customization in the service's source code.

Applications don't (currently) have that same requirement.

Actually, this probably warrants a discussion...

apps/obc-hs/README.rst
cantino: Thanks for the explanation!

apps/obc-hs/README.rst
plauche: If this is just an example, then I think it's ok for it to require modification to be usable. However if this is intended to be a more generic, reusable application, then that's a different story. Which is the intent with this application?

apps/obc-hs/src/check_mem.rs
plauche: Is `/tmp` factored into this calculation?

apps/obc-hs/README.rst
```diff
@@ -0,0 +1,111 @@
...
+in use and checks them against the given threshold values (nominal, high, critical).
+
+If the RAM usage percentage exceeds the critical threshold (default: 80%), then the system is
+forcibly rebooted.
+The assumption in this case is that a rogue process has an undiscovered memory leak.
+As a result, rebooting the system will free the memory back up.
+
+If the disk usage percentage exceeds the critical threshold (default: 80%), then the ``clean_db``
+function is called again, however the threshold age is lowered to a smaller value (default: 1 day).
+Reducing the amount of storage used by the telemetry database is the only common way of reducing
```
plauche: What about deleting old log files?

apps/obc-hs/src/check_fs.rs
```diff
@@ -0,0 +1,84 @@
...
+    // Check that the user data partition is still writeable.
+    let test_file = "/home/obc-test-file";
+    let test_string = "I'm a test string";
+
+    if check_write(test_file, test_string)
+        .and(check_read(test_file, test_string))
+        .is_err()
+        && COMMS_SERVICE != ""
+    {
+        let _ = send_error();
```
plauche: Is this the only condition that we want reported to the comms service

apps/obc-hs/src/check_mem.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    let result = query(&service, OBC_TELEMETRY, Some(QUERY_TIMEOUT))?;
+
+    let mem = result["memInfo"]["available"].as_u64().unwrap_or_else(|| {
+        error!("Failed to fetch available memory");
+        0
+    });
+
+    // Convert to percentage in use, since that's an easier number to work with
+    let ram_in_use = (100 - mem * 100 / MEM_TOTAL) as u8;
```
plauche: Can we use `result["meminfo"]["total"]` for our total amount of system ram?

apps/obc-hs/src/check_mem.rs
```diff
@@ -0,0 +1,115 @@
...
+    } else {
+        error!("RAM usage too high: {}%. Triggering reboot", ram_in_use);
+        // Reboot the system not-nicely. If we're at this point, there's probably a rogue process
+        // that's hogging all the system resources and not playing nicely with others.
+        Command::new("reboot").arg("-f").status()?;
+    }
+
+    // Check disk space usage
+    // Get the % of the user data partition that's free
+    let disk_in_use = if let Ok(output) = Command::new("df").arg(USER_PARTITION).output() {
```
plauche: Do we have `du` on the system? That tool could allow us to pinpoint disk usage to the telem db vs copies of apps vs other stuff.

apps/obc-hs/README.rst
cantino: If it's an example, should it be in a different part of the repo that indicates is should be copied, not sourced?

apps/obc-hs/README.rst
CatGarab: You could, however the log files are already configured with a max size (all added up, it's 2.4MB). I don't have super strong feelings, but:
- They're unlikely to be the memory-eating culprit
- Log files are one of the main ways to debug a problem like this
- Deleting the older log files won't free up that much memory

apps/obc-hs/src/check_fs.rs
CatGarab: probably? the original intention was that a separate beacon application would be responsible for generating the various radio beacon packets. This includes an error packet with the current errors count, as pulled from the errors log.

If we hit this line, we've lost the ability to write to the user data partition, which means we're no longer adding errors to the errors log, so the usual errors beacon becomes non-functional. As a result, this line is only way to detect and notify the ground of the issue

apps/obc-hs/src/check_mem.rs
CatGarab: `du` is enabled on the system. however, i don't think there's much automated logic which we could add here to assist with that. Running `du` and looking for data-hogs seems like a manual process that is suited to the shell service

apps/obc-hs/src/check_mem.rs
CatGarab: oh right, this check only looks at the data usage in the user data partition. `/tmp` is therefore unrelated and untracked

PR #359, opened 2019-05-23
Updating Hardware Service Required Schema

jacoffey3:
Incorporating some of the lessons learned from implementing a few hardware services in house: 

1. We overdefined them. A lot of the functionality I set forth didn't make sense when it hit hardware, and had to have interesting workarounds to implement
1. We didn't use many of them. For the actual housekeeping/monitoring/running the spacecraft, many of the mutations and queries were just unused. 

So, with this PR, I'm proposing we strip it down to just a few essential things we want to require be implemented: 

- Queries: 
  1. ping: For testing that the service is up and running regardless of hardware
  1. config: Quick reference to software version of the hardware, how the device is configured, whatever makes sense for the service. This allows us to grab it as a top level item for us to downlink specially. 
  1. power: Pretty self explanatory
  1. telemetry: This is the same, but we're getting rid of the nominal/debug categories, since they were actually not useful when interacting with the service or doing automatic collection. I'm going to open up a separate issue discussing how to automate collection of telemetry from hardware services. 
- Mutations: 
  1. noop: For testing that the hardware is able to communicated with. Allows us to automate housekeeping. 
  1. controlPower: Pretty self explanatory
  1. commandRaw: We want a way to always have access to the entire capability the hardware provides, in case we experience an edge case that isn't taken care of by the mutations we want to expose.


jacoffey3:
[Update hardware service schema docs](https://trello.com/c/r4VX1AIk/414-update-hardware-service-schema-docs)

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,78 @@
...
+This guide covers an outline for the base Schema that should be implemented in all hardware services.
 For general information about hardware services, their role, and how they work, check out :doc:`the hardware services documentation. <../services/hardware-services>`
-Make sure you reference the general hardware service when looking at the schema for all others, because they simply build on that schema, and each section covers what is added to that base schema.
-
-
-General Hardware Service
-------------------------
 
-A general hardware service is a service for any piece of hardware that does not fit into any of the other categories. These queries/mutations will be expected to be present regardless of the hardware. All other service outlines build on top of what is present here.
+All hardware services must follow the graphQL spec.
```
cantino: It's *GraphQL*.

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,78 @@
 KubOS Service Outlines
 ======================
 
-TODO: Review this. Is any of it outdated?
-
-This guide covers the development of KubOS hardware services and provides an outline for several major types of hardware.
+This guide covers an outline for the base Schema that should be implemented in all hardware services.
```
cantino: Maybe add: "Your service can implement more mutations and fields beyond those specified in the required base." or something.

jacoffey3:
> Once we decide on automating the collecting, I assume it may result in an additional required field with the fragment or list of recommended fields.

I'm going to make another issue for that, I think it will most likely be in another doc.

jacoffey3:
https://github.com/kubos/kubos/issues/364

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,79 @@
...
+The mutation and query responses are the "data" and "errors" fields, and all errors experienced are returned in that "errors" field, even if they are from the underlying hardware.
 
-GraphQL Schema::
+.. code-block:: graphql
 
     type Query {
-        ack(): String
+        ping(): String
+        config(): ConfigObject
         power(): PowerState
```
plauche: Does `power` just check whether the system thinks power is applied to the hardware device, or does it actually confirm that the device is powered on and responsive?

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,79 @@
...
-        # results1: any type
-        # results2: any type
-        # ...
-    }
 
     type Mutation {
-        noop(): NoopPayload
+        # When the service receives a noop mutation, it should send a command and make sure it sees the appropriate response from the hardware.
+        noop(): Boolean
         controlPower(
```
plauche: How do we handle this in conditions where another service is actually responsible for the power switch? Would this just trigger whatever power api is exposed by this hardware, or would it defer to another (eps) service?

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,79 @@
...
 
     type Query {
-        ack(): String
+        ping(): String
+        config(): ConfigObject
         power(): PowerState
-        config(): String
-        errors(): [String] # Error descriptions if there are any, or empty if there aren't
         telemetry(): Telemetry
-        testResults(): TestResults
```
plauche: Should we also have a `queryRaw` or `telemetryRaw` field for any informational items which aren't exposed in the schema but we still might want access to?

docs/os-docs/services/service-outline-guide.rst
CatGarab: Should confirm that it's powered on and responsive. Most devices have an `uptime` field which can be queried to verify this

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,79 @@
...
 For general information about hardware services, their role, and how they work, check out :doc:`the hardware services documentation. <../services/hardware-services>`
-Make sure you reference the general hardware service when looking at the schema for all others, because they simply build on that schema, and each section covers what is added to that base schema.
-
-
-General Hardware Service
-------------------------
 
-A general hardware service is a service for any piece of hardware that does not fit into any of the other categories. These queries/mutations will be expected to be present regardless of the hardware. All other service outlines build on top of what is present here.
+All hardware services must follow the GraphQL spec.
+The mutation and query responses are the "data" and "errors" fields, and all errors experienced are returned in that "errors" field, even if they are from the underlying hardware.
```
CatGarab: Could use an example of this

docs/os-docs/services/service-outline-guide.rst
CatGarab: - Most devices have a software reset command available. We'd still want to expose that here. The hardware reset that you're talking about should just be deferred to whatever service has control over it
- In the case where there isn't a power command of any sort, we've still had this mutation, but just returned a "Not Implemented" string ([example](https://github.com/kubos/kubos/blob/master/services/novatel-oem6-service/src/schema.rs#L242))
    - This behavior should definitely be documented, though

docs/os-docs/services/service-outline-guide.rst
CatGarab: The `commandRaw` mutation will support returning a response. That would cover this case, so we don't necessarily need a separate query.
(All of our current services need to be updated to support this, though...)

CatGarab:
The `commandRaw` mutation, in particular, could use some in-depth docs to explain how the different data formats (hex vs string) work

docs/os-docs/services/service-outline-guide.rst
jacoffey3: ☝️ That. Yeah, this will be a fairly major update.

docs/os-docs/services/service-outline-guide.rst
jacoffey3: Many devices will have some sort of power control available that is specific to the device itself. If it doesn't, as @CatGarab said, the "Not Implemented" string should be returned in the errors field.

I mainly want to make sure they will expose it in this way _if_ it is available.

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,79 @@
...
+Your service can and should implement more mutations and queries beyond those specified in the required base.
 For general information about hardware services, their role, and how they work, check out :doc:`the hardware services documentation. <../services/hardware-services>`
-Make sure you reference the general hardware service when looking at the schema for all others, because they simply build on that schema, and each section covers what is added to that base schema.
-
-
-General Hardware Service
-------------------------
 
-A general hardware service is a service for any piece of hardware that does not fit into any of the other categories. These queries/mutations will be expected to be present regardless of the hardware. All other service outlines build on top of what is present here.
+All hardware services must follow the GraphQL spec.
```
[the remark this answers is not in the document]
jacoffey3: No I need to link the spec.

jacoffey3:
> Since there aren't that many required things, it would be nice if the schema was broken out into a more doc-ish format with more in-depth descriptions. Ie have a "Queries" doc section and a "Mutations" doc section and then have each available request as a sub-section.

I think that's a good idea. I'll break it out into the sections and move the comments I've put in to those sections. I still want to keep the schema itself though.

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,18 +1,34 @@
-KubOS Service Outlines
+KubOS Service Outline
 ======================
```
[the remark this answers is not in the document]
jacoffey3: ....seriously ryan

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -68,11 +93,20 @@ The mutation and query responses are the "data" and "errors" fields, and all err
         # Necessary for binary data to be passed to the hardware
```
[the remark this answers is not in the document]
jacoffey3: Yeah I can reword that

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,112 @@
...
 
-TODO: Review this. Is any of it outdated?
-
-This guide covers the development of KubOS hardware services and provides an outline for several major types of hardware.
+This guide covers an outline for the base Schema that should be implemented in all hardware services.
+Your service can and should implement more mutations and queries beyond those specified in the required base.
 For general information about hardware services, their role, and how they work, check out :doc:`the hardware services documentation. <../services/hardware-services>`
-Make sure you reference the general hardware service when looking at the schema for all others, because they simply build on that schema, and each section covers what is added to that base schema.
 
+All hardware services must follow the :url:`GraphQL spec. <https://graphql.github.io/graphql-spec/>`
```
plauche: This url syntax is not correct.
```suggestion
All hardware services must follow the `GraphQL spec. <https://graphql.github.io/graphql-spec/>`__.
```

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,110 @@
...
+_______________________
 
-If there are any major service categories that should be added, or if you feel that any sections are missing Queries or Mutations that would be essential for Mission Applications, please let us know on our `Slack <https://slack.kubos.co/>`__ or open a PR to add them yourself!
+Direct hardware commands are often issued in binary format, which GraphQL does not have a built in type for.
+Since we want to support the passing of raw binary commands to underlying hardware, but do not want to deviate from the GraphQL specification, we've added the FormatEnum to specify how the utf-8 compatible string can be decoded to the raw data that must be passed to the hardware.
+The ``HEX`` format is for those cases, such that passing a hex string: ``74657374636f6d6d616e64``, causes the bytearray: ``[116,101,115,116,99,111,109,109,97,110,100]`` to be passed to the hardware.
 
+Some hardware expects string commands, so those services would leverage the ``STR`` format to pass the data directly to the hardware, with no decode necessary. EG: the string ``testcommand`` gets passed to the hardware as the bytearray: ``[116,101,115,116,99,111,109,109,97,110,100]``.
```
CatGarab: "The `STR` format declares that the input string consists of UTF-8 characters. In this case, the string {...}"

docs/os-docs/services/service-outline-guide.rst
```diff
@@ -1,377 +1,110 @@
...
 
-Additional Services
--------------------
+``commandRaw`` mutation
+_______________________
 
-If there are any major service categories that should be added, or if you feel that any sections are missing Queries or Mutations that would be essential for Mission Applications, please let us know on our `Slack <https://slack.kubos.co/>`__ or open a PR to add them yourself!
+Direct hardware commands are often issued in binary format, which GraphQL does not have a built in type for.
+Since we want to support the passing of raw binary commands to underlying hardware, but do not want to deviate from the GraphQL specification, we've added the FormatEnum to specify how the utf-8 compatible string can be decoded to the raw data that must be passed to the hardware.
+The ``HEX`` format is for those cases, such that passing a hex string: ``74657374636f6d6d616e64``, causes the bytearray: ``[116,101,115,116,99,111,109,109,97,110,100]`` to be passed to the hardware.
```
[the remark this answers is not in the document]
jacoffey3: I don't understand what you mean.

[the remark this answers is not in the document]
CatGarab: i was being lazy and didn't want to type out the remainder of the sentence. The `{...}` translates to "blah blah blah", which translates to basically just using what you have for the rest of the sentence.
So:
The HEX format declares that the input string is already in hexadecimal format. In this case, passing hex string `74657374636f6d6d616e64` causes the byte array `[116,101,115,116,99,111,109,109,97,110,100]` to be passed to the hardware.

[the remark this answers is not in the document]
jacoffey3: Got it. That's essentially what I have there now.

PR #362, opened 2019-05-23
v1.15 changelog

CatGarab:
Adding changelog notes for v1.15

Also updating a lot of the old links to point to their particular version of the docs, rather than `latest`. Doing this because the upcoming doc re-org will invalidate a lot of the links and I'd like for us to still point to something, rather than having a bunch of 404s


CatGarab:
[1.15 changelog](https://trello.com/c/nnq5lm9N/908-115-changelog)

PR #368, opened 2019-05-29
Fixing intermittently failing CI tests

CatGarab:
We have a few unit tests that fail periodically. This PR attempts to fix them

- Updating boot vars tests to use a randomly generated temporary directory rather than just `/tmp`
- Updating boot var tests to add a little time between setting a variable and then fetching it
- Expanding the delay for the iMTQ watchdog tests
- Making all CBOR doc examples `no_run`
- Tweaking the way we change the current directory for running apps
- Adding retry logic to the app service test setup

### DIFF
--- a/tools/ci_c.py
+++ b/tools/ci_c.py
@@ -43,6 +43,7 @@ def build(dir):
 
 def run_test(dir):
     build_dir = "build"
+    os.environ["CTEST_OUTPUT_ON_FAILURE"] = "1"
     subprocess.run(["make", "test"], cwd=build_dir, check=True)
 
 def test(dir):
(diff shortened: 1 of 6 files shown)


CatGarab:
[Fix unit tests for CI](https://trello.com/c/5jN7P97i/756-fix-unit-tests-for-ci)

tools/ci_c.py
```diff
@@ -43,6 +43,7 @@ def build(dir):
 
 def run_test(dir):
     build_dir = "build"
+    os.environ["CTEST_OUTPUT_ON_FAILURE"] = "1"
```
plauche: Does this actually fix a test or is this for debugging?

CatGarab: This is for debugging, however I'd like to leave it in. It only prints out extra stuff when something fails and is required in order to find out what the actual failure is

PR #369, opened 2019-05-29
Initial implementation of insertBulk for the telemetry db service

marshall:
This initial version can insert multiple entries with an optional timestamp per-entry, an optional top level timestamp, or using system time (the default behavior). Included an initial unit test

/cc @plauche @CatGarab for 👀  and feedback on the API


marshall:
[Bulk Inserts for Telem DB service to improve input performance](https://trello.com/c/hoeATk0B/707-bulk-inserts-for-telem-db-service-to-improve-input-performance)

apis/telemetry-db-api/src/models.rs
```diff
@@ -17,19 +17,11 @@
...
+#[derive(Debug, Queryable, Serialize, Deserialize, Insertable)]
+#[table_name = "telemetry"]
 pub struct Entry {
     pub timestamp: f64,
     pub subsystem: String,
     pub parameter: String,
     pub value: String,
 }
-
-#[derive(Insertable)]
```
plauche: :+1: Nice that we could collapse this into one model.

services/telemetry-service/src/main.rs
```diff
@@ -204,6 +204,7 @@
 
 #[macro_use]
 extern crate juniper;
+extern crate time;
```
CatGarab: nit: we're using Rust 2018 edition, which no longer requires that crates be declared like this

services/telemetry-service/tests/test_insert.rs
```diff
@@ -269,3 +269,69 @@ fn test_insert_current_timestamp() {
     let difference = (now - timestamp).abs();
     assert!(difference < 1.0);
 }
+
+#[test]
+fn test_insert_bulk() {
```
CatGarab: Need more unit tests:
- Inserting with a top-level timestamp
- Inserting where at least one entry has its own timestamp

services/telemetry-service/src/main.rs
marshall: I'm elated to find out this is a thing

marshall:
Updated with docs, and an update to the telemetry DB benchmark to include both API and GQL bulk insert.

Benchmark results for me locally on a Macbook Pro 6-core 2.2GHz i7 w/ 32GB RAM:
```
+----------------------------+------------------+----------+
| NAME                       | Avg us per-entry | Total us |
+----------------------------+------------------+----------+
| DB API insert (local)      | 541              | 541815   |
+----------------------------+------------------+----------+
| DB API insert_bulk (local) | 39               | 39005    |
+----------------------------+------------------+----------+
| GQL insert (remote)        | 2570             | 2570391  |
+----------------------------+------------------+----------+
| GQL insertBulk (remote)    | 354              | 354878   |
+----------------------------+------------------+----------+
| UDP insert (remote)        | 69               | 69865    |
+----------------------------+------------------+----------+
```

PR #370, opened 2019-05-29
Adding udp passthrough to comms service.

plauche:
UDP pass through is implemented one way from ground to flight services. We are currently assuming that any responses will be passed through the downlink port.


plauche:
[UDP passthrough for Comms Service](https://trello.com/c/JxI44SHM/689-udp-passthrough-for-comms-service)

plauche:
@CatGarab agreed about updating the doc, do you mind if I split that into a new task (https://trello.com/c/IRYZz7iS/932-update-comms-service-doc-to-reflect-new-udp-passthrough-method) and address that in the docs re-org effort, rather than intro docs merge conflicts in master?

PR #372, opened 2019-05-30
Modifying file-service to allow configurable downlink ip/port

plauche:


plauche:
[File service: Non-default destination IP/port](https://trello.com/c/DWWqPJr0/693-file-service-non-default-destination-ip-port)

CatGarab:
Why did all the multi-client tests get removed?

CatGarab:
What's the benefit of forcing the user to configure `downlink_port`?

clients/kubos-file-client/src/main.rs
```diff
@@ -164,10 +183,16 @@ fn main() {
...
-                .short("h")
+                .short("-h")
                 .takes_value(true)
                 .default_value("0.0.0.0"),
         )
+        .arg(
+            Arg::with_name("host_port")
+                .short("-P")
+                .takes_value(true)
+                .default_value("8000"),
```
CatGarab: If the service is going to require the user to manually define the `downlink_port` value, then it should also be a required arg here

docs/services/file.rst
```diff
@@ -169,6 +169,8 @@ defined in the system's :doc:`config.toml <../services/service-config>` file:
           in bytes.
         - ``hold_count`` - `Default: 5.` The number of times the protocol waits for
           a new message before ending the transaction.
+        - ``downlink_ip`` - `Default: 127.0.0.1` The IP address that the file service responds to.
+        - ``downlink_port`` - `Required` The port that the file service responds to.
```
CatGarab: I'd prefer that `downlink_ip` and `downlink_port` either both be required or both be optional

services/file-service/src/lib.rs
```diff
@@ -50,6 +50,14 @@ pub fn recv_loop(config: &ServiceConfig) -> Result<(), failure::Error> {
         None => 5,
     } as u16;
 
+    let downlink_port = config.get("downlink_port").unwrap().as_integer().unwrap() as u16;
```
CatGarab: I'm okay with this panicking, but throw a descriptive error so that it's easy to see the problem

services/file-service/src/lib.rs
```diff
@@ -93,17 +101,24 @@ pub fn recv_loop(config: &ServiceConfig) -> Result<(), failure::Error> {
             let (sender, receiver): (Sender<serde_cbor::Value>, Receiver<serde_cbor::Value>) =
                 mpsc::channel();
             threads.lock().unwrap().insert(channel_id, sender.clone());
+
             // Break the processing work off into its own thread so we can
             // listen for requests from other clients
             let shared_threads = threads.clone();
+            let downlink_ip_ref = downlink_ip.to_owned();
```
CatGarab: It would be more convenient if the downlink IP was derived from `source` (if `downlink_ip` stays optional)

CatGarab:
I got the file client to work with my VM, but I had to update my Vagrantfile in order to expose my machine with a [public network](https://www.vagrantup.com/docs/networking/public_network.html#static-ip). Without it, my host machine was doing dynamic port forwarding...or something...so the host port that I defined for the file client wasn't the actual port being used.

The line I added looks like this:
```
config.vm.network "public_network", ip: "168.168.2.30", bridge: "Intel(R) Ethernet Connection I219-V"
```

Where `168.168.2.*` is the network I defined for my BBB and ethernet adapter

plauche:
@CatGarab to answer both of your questions - This change is made to make the file service functional again over the communications service. Under the current design, Udp packets sent from ground -> flight are handled in individual, temporary passthrough threads. Udp packets sent from flight -> ground are all passed through the comm service's UDP downlink port.

This design requires the file service to have the ability to send data to a fixed port, while receiving data on a separate fixed port, hence the configuring of `downlink_port`.

The multi-client tests went away because it was too much trouble to create a test environment where multiple clients could co-exist on the same system while listening for data on the same port. This limitation would not exist in the real-life scenario of multiple ground stations, as the separate systems could all have one file client listening on the same port.

services/file-service/src/lib.rs
```diff
@@ -50,6 +50,23 @@ pub fn recv_loop(config: &ServiceConfig) -> Result<(), failure::Error> {
...
+
+    let downlink_ip = config
+        .get("downlink_ip")
+        .expect("Downlink IP not found")
+        .as_str()
+        .and_then(|str| Some(str.to_owned()))
+        .expect("Downlink IP not found");
+
+    info!(
+        "Starting file transfer service at {}:{}",
```
CatGarab: Break this into two messages:
- The IP and port the file service is listening on
- The IP and port the file service is downlinking to

clients/kubos-file-client/README.rst
```diff
@@ -21,6 +21,7 @@ Required arguments:
         - ``download`` - Transfer ``source-file`` on the remote target to ``target-file`` location
                        on the local host
     - ``source-file`` - The file to be transferred. May be a relative or absolute path.
+    - ``P {host_port}`` - The UDP port that the file transfer service will send responses to.
```
CatGarab: nit: `-P`

docs/tutorials/file-transfer.rst
```diff
@@ -50,6 +50,7 @@ Optional arguments:
       If not specified, the root file name from ``source-file`` will be used and the file will be
       placed in the current directory of the destination.
     - ``-h {host IP}`` - Default: `0.0.0.0`. IP address of the local host to use.
+    - ``-P {host port}`` - Default: `8000`. UDP port of the local host to listen on.
```
CatGarab: This arg is required, right?

You'll also need to update the example commands in this tutorial to add the `-P` argument to all of them

PR #381, opened 2019-06-06
Locking version of breathe until we fully migrate to Python3

CatGarab:
When I tried to regenerate our Docker image, it failed because the latest version of `breathe` only supports Python3. I have locked it at the last Python2.7-friendly version and created a couple cards to migrate everything to Python3 in the future.

There are two other lines which I added while messing with things that I would like to leave in:
- Upgrading pip to the latest version - This is just because the default version is super old
- Upgrade setuptools to the latest version - I ran into issues when installing `sphinxcontrib-plantuml` for Python3. It threw a really weird (non-obvious) error. A google search led me to conclude that `setuptools` needed to be updated.

### DIFF
--- a/tools/dist/Dockerfile
+++ b/tools/dist/Dockerfile
@@ -1,4 +1,4 @@
-# kubos/kubos-dev:1.15.0
+# kubos/kubos-dev:1.15.1
 
 FROM phusion/baseimage:0.9.22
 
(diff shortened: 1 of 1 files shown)


CatGarab:
[Update xcompile toolchains](https://trello.com/c/UxUdvM2N/941-update-xcompile-toolchains)

CatGarab:
[Update xcompile toolchains](https://trello.com/c/UxUdvM2N/941-update-xcompile-toolchains)

PR #382, opened 2019-06-06
Removing problematic Docker setup commands

CatGarab:


CatGarab:
[Update xcompile toolchains](https://trello.com/c/UxUdvM2N/941-update-xcompile-toolchains)

PR #389, opened 2019-06-17
updated cross-compile docs to use `cargo kubos`

marshall:
depends on kubos/cargo-kubos#4


marshall:
[Build auto-CC into SDK](https://trello.com/c/AvlNhXp1/772-build-auto-cc-into-sdk)

PR #394, opened 2019-06-20
Adding instructions for local cross-compile

CatGarab:
Adding instructions for setting up cross-compiling in a local dev environment, rather than the SDK.

The SDK docs already have some information about cross-compiling, but I'm leaving that where it is for now. Once the dust settles (or if someone has a better idea), the information might get moved around.

### DIFF
--- a/docs/obc-docs/cross-compile.rst
+++ b/docs/obc-docs/cross-compile.rst
@@ -2,7 +2,7 @@ Compiling Code for a Target OBC
 ===============================
 
 In order for programs written using a compiled language (Rust, C, etc) to work on a target OBC,
-it will need to be compiled using the appropriate cross-compile toolchain.
+they will need to be compiled using the appropriate cross-compile toolchain.
 
 This document covers the process of installing the cross-compile toolchain and compiling a program
 for a target OBC using a local development environment.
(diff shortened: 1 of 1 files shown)


CatGarab:
[Cross-compile doc](https://trello.com/c/xZhUqGgz/888-cross-compile-doc)

docs/obc-docs/cross-compile.rst
```diff
@@ -1,8 +1,80 @@
 Compiling Code for a Target OBC
 ===============================
 
-TODO: Rust and C need to use our toolchains
+In order for programs written using a compiled language (Rust, C, etc) to work on a target OBC,
```
cantino: *they* will

docs/obc-docs/cross-compile.rst
```diff
@@ -1,8 +1,80 @@
...
+for reference.
+You should be able to directly copy the lines pertaining to your target OBC.
+
+Using the Toolchains
+--------------------
+
+Rust
+~~~~
+
+The new toolchains can be used by specifying the target triple you defined in the ``--target``
```
cantino: Maybe clearer: 

```
The new toolchains can be used by specifying the previously-defined target triple in the ``--target`` argument when running ``cargo build``.
```

docs/obc-docs/cross-compile.rst
```diff
@@ -1,8 +1,80 @@
...
\ No newline at end of file
+The Kubos SDK comes with the cross-compile toolchains pre-installed, as well as some additional
+helper tooling.
+If you are using the SDK as your development environment, please refer to :doc:`those docs <../sdk-docs/index>`
+for more specific information.
+
+Download
+--------
+
+Below are the direct download links for the latest cross-compiling toolchains.
```
cantino: Worth explaining what is actually in these toolchain directories?

docs/obc-docs/cross-compile.rst
```diff
@@ -1,8 +1,80 @@
...
+
+The new toolchains can be used by specifying the target triple you defined in the ``--target``
+argument when running ``cargo build``.
+
+For example::
+
+    $ cargo build --target arm-unknown-linux-gnueabihf
+    
+More specific information about how to set up and use a Rust project can be found in the
+:doc:`../sdk-docs/sdk-rust` doc.
```
jacoffey3: Wouldn't this be a link to the getting started with rust section? Or is that for dependencies? Seems like a weird place to link to.

docs/obc-docs/cross-compile.rst
CatGarab: Thanks! That sentence was definitely giving me problems

docs/obc-docs/cross-compile.rst
CatGarab: oh right, i forgot about that one

CatGarab:
@plauche Ultimately yes, we should have instructions for installing `cargo kubos`, however I'd like to do that work in a different PR since the current `cargo kubos` PRs haven't landed yet

I created a card: https://trello.com/c/izDqi3mI/979-document-how-to-install-and-configure-cargo-kubos

PR #395, opened 2019-06-21
Updating the windows dev guide

CatGarab:
Turns out Windows users can now develop out of a folder that's shared with the SDK like everyone else. As a result, I have removed the old "remote development environment" instructions.


CatGarab:
[Getting Started - Windows Dev Environment](https://trello.com/c/TyLwwhaw/952-getting-started-windows-dev-environment)

docs/getting-started/windows-setup.rst
```diff
@@ -1,10 +1,107 @@
 Setting up a Dev Environment in Windows
 =======================================
 
-1. Cry
-2. Cry some more
-3. :ref:`PowerShell <powershell>`
-4. Set up a VM
+Developing for a Linux-based target system from within a Windows-based host machine is difficult.
+The way Windows has architected its file systems and behaviors is fundamentally different from
+Linux, making it impossible to run KubOS services and applications locally.
```
jacoffey3: ~~impossible~~ extremely difficult
I'm guessing it's very possible, just not something we are going to support anytime soon.

docs/getting-started/windows-setup.rst
```diff
@@ -1,10 +1,107 @@
...
+re-built.
+
+Additionally, developing out of a shared folder allows you to use your normal IDE for project
+development.
+You will just need to execute any commands related to building and running the project from the
+SDK's command line.
+
+See the instructions on how to :ref:`mount a host directory <mount-directory>` in your VM for more
+information.
+
```
jacoffey3: I think you should reference them somewhere at the end of this. Maybe getting started/installing the SDK?

CatGarab: Sorry, I'm not following. Reference what where?

docs/getting-started/windows-setup.rst
plauche: I think the bigger issue would be getting the services compiled for Windows. Folder structure shouldn't be as much of an issue, since we can point to a custom `config.toml`.

CatGarab: We currently have various references to things like UnixStreams and UnixDatagrams in our dependencies, which prevents things from being built for a Windows target

docs/getting-started/windows-setup.rst
```diff
@@ -1,10 +1,119 @@
...
+11. In the "When does this rule apply?" menu, leave all checkboxes selected
+
+.. image:: ../images/inbound_rule_network.png
+
+12. In the "Name" field, enter something descriptive for the rule. For example, "Allow UDP from OBC".
+    Then click "Finish" to finalize and activate the new rule.
+
+.. image:: ../images/inbound_rule_name.png
+
+Next Steps
```
jacoffey3: 🎉

PR #396, opened 2019-06-21
Remove undefined and unused module from test

cantino:
`app-service` tests failed with `ModuleNotFoundError: No module named 'app_api'` until I removed this. If we want to use it, I think we'd need to import it from `apis/app-api/python`. How was this intended to be loaded?

### DIFF
--- a/services/app-service/tests/utils/python-proj/main.py
+++ b/services/app-service/tests/utils/python-proj/main.py
@@ -4,7 +4,6 @@
 # Licensed under the Apache License, Version 2.0
 # See LICENSE file for details.
 
-import app_api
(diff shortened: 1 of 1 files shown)


CatGarab:
The app api is loaded into the [CI docker image](https://github.com/kubos/kubos/blob/master/.circleci/config.yml#L48) prior to running the unit tests. For users using the SDK, it should already be installed as a [Python3 library](https://github.com/kubos/kubos-vagrant/blob/master/kubos-dev/script/non-privileged-provision.sh)

cantino:
Okay, thanks!

PR #403, opened 2019-06-26
Getting Started with Rust

plauche:


docs/getting-started/using-rust.rst
```diff
@@ -1,7 +1,41 @@
...
+The binary can be run locally using the ``cargo run`` command.
+Any desired arguments can be passed to the underlying executable by placing them behind ``--`` like
+so::
+
+    $ cargo run -- -c config.toml
+
+Next Steps
+----------
+
+Ready to create a mission application in Rust? Take a look at the :doc:`rust mission application <../tutorials/first-rust-app>` tutorial.
```
plauche: Should this `getting started with rust` go into building a mission app? Should we create new tutorials which cover making a rust mission app and rust service?

plauche:
[Getting Started with KubOS and Rust](https://trello.com/c/VgrcKBpK/954-getting-started-with-kubos-and-rust)

docs/getting-started/using-rust.rst
CatGarab: yes and yes.
I think you should point to the current first-app tutorial for now (and maybe mention that it's in python and should just be used as a reference)
And then we should add Rust-based tutorials at some point in the future

docs/getting-started/using-rust.rst
```diff
@@ -1,7 +1,41 @@
...
+installed prior to using Rust.
 
-All KubOS dependencies can be referred to by specifying the top-level git repo url in the Cargo.toml
-file.
+New Project
+-----------
+
+A new Rust project can be created by running the following command:
+
+  cargo new project_name
```
cantino: Is this something a user should run when starting out? Where should they execute it?

docs/getting-started/using-rust.rst
```diff
@@ -1,7 +1,41 @@
 Getting Started with KubOS and Rust
 ===================================
 
-This is **only** for running rust locally. No cross-compiling yet.
+This document will cover getting started using Rust for local KubOS development.
+You will need to have the build dependencies listed in :ref:`<build-dependencies>`
```
CatGarab: nit: ref formatting. You shouldn't wrap the reference with the wakkas if you're not providing an alternate title
```
:ref:`build-dependencies`
:ref:`Alt Title <build-dependencies>`

File: docs/getting-started/using-rust.rst [hunk 4]
```diff
@@ -1,7 +1,41 @@
...
+installed prior to using Rust.
 
-All KubOS dependencies can be referred to by specifying the top-level git repo url in the Cargo.toml
-file.
+New Project
+-----------
+
+A new Rust project can be created by running the following command:
```
CatGarab: nit: need a second colon to make the next line a `code block`

docs/getting-started/using-rust.rst
CatGarab: nit: Pre-pend with `$` like the other example commands

docs/getting-started/using-rust.rst
jacoffey3: > Should this `getting started with rust` go into building a mission app? Should we create new tutorials which cover making a rust mission app and rust service?

It should go into it, but not with this reorg. Make a card in the backlog to add it to this doc later.

docs/getting-started/using-rust.rst
```diff
@@ -1,7 +1,50 @@
...
+Next Steps
+----------
+
+Ready to create a mission application in Rust? Take a look at the :doc:`Python mission application <../tutorials/first-mission-app>`
+tutorial for guidance in the construction of a mission application.
+The `example rust mission application <https://github.com/kubos/kubos/tree/master/examples/rust-mission-app>`__ is a good model for a Rust-based mission application.
+
+
+Interested in cross-compiling your project for use on hardware? Take a look at the
+:doc:`local cross-compiling <local-cross-compiling>` guide.
```
CatGarab: The cross compiling doc is in `obc-docs/cross-compile`

PR #406, opened 2019-07-01
Renaming 'os-docs' folder to 'ecosystem'

CatGarab:


CatGarab:
@plauche I updated the changelog in a prior PR (outside of the doc reorg) and version locked all of the doc links so that they'll continue to be valid

PR #411, opened 2019-07-02
Updating Rust app API and example apps to allow custom config file

CatGarab:
- Updated the rust app api to allow the `-c config` option
- Updated the rust system-api to allow the `-r level` option (not that it will be used by that crate, but it was the easiest/fastest way to fix this issue. The `getopt` crate does not have a way to allow unknown args. We might want to migrate to something like `clap` in the future)
- Updated the python examples to get and set the non-default config value

Doc updates will be done in a separate PR (against the doc-reorg branch)

### DIFF
--- a/apis/app-api/rust/src/framework.rs
+++ b/apis/app-api/rust/src/framework.rs
@@ -149,6 +149,13 @@ pub fn app_start(
         "Run level which should be executed",
         "RUN_LEVEL",
     );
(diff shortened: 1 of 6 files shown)


jacoffey3:
[Can't pass custom config file to a rust app or a python app](https://trello.com/c/NhcVpBhu/903-cant-pass-custom-config-file-to-a-rust-app-or-a-python-app)

PR #412, opened 2019-07-02
Getting started python doc

plauche:


plauche:
[Getting Started with KubOS and Python](https://trello.com/c/VWCOnZq3/955-getting-started-with-kubos-and-python)

PR #413, opened 2019-07-03
Merging master into doc-reorg

plauche:


plauche:
I *think* we should only need to check any changed `.md` or `.rst` files in this merge to see if any new content needs to get shifted around. The actual merge conflicts were pretty minimal and easily resolved.

PR #416, opened 2019-07-03
Fixing broken docs links

plauche:


plauche:
There are still two broken links in `mission-needs.rst` which attempt to reference the housekeeping app from #358. Do we want to land that PR or just remove the references?

jacoffey3:
> There are still two broken links in `mission-needs.rst` which attempt to reference the housekeeping app from #358. Do we want to land that PR or just remove the references?

Let's not block on the PR, but I'll add a note to that PR. I'll reword that sentence.

jacoffey3:
I suggested changes in the base PR to fix it. Should've done it here, but forgot...sorry.

PR #417, opened 2019-07-03
Minor cleanup on first-mission-app

cantino:
Fix some paths and move `SERVICES = app_api.services()` into the conditional because otherwise it crashes with `No such file or directory: /home/system/etc/config.toml` (a default I assume).

A deeper issue is that I cannot use use the example app because `mem_info` does not work on a Mac (no `proc` filesystem). This script does work, albeit with fake data:

```python
#!/usr/bin/env python3

import argparse
import app_api
import sys


def on_boot():

    print("OnBoot logic")

def on_command():

    request = '{ping}'

    try:
        response = SERVICES.query(service="monitor-service", query=request)
    except Exception as e:
        print("Something went wrong: " + str(e))
        sys.exit(1)

    data = response["ping"]

    print("Response: %s" % (data))

    request = '''
        mutation {
            insert(subsystem: "OBC", parameter: "available_mem", value: "%s") {
                success,
                errors
            }
        }
        ''' % (1234)

    try:
        response = SERVICES.query(service="telemetry-service", query=request)
    except Exception as e:
        print("Something went wrong: " + str(e) )
        sys.exit(1)

    data = response["insert"]
    success = data["success"]
    errors = data["errors"]

    if success == False:
        print("Telemetry insert encountered errors: " + str(errors))
        sys.exit(1)
    else:
        print("Telemetry insert completed successfully")

def main():

    parser = argparse.ArgumentParser()

    parser.add_argument('--run', '-r')
    parser.add_argument('--config', '-c')

    args = parser.parse_args()

    if args.config is not None:
        global SERVICES
        SERVICES = app_api.Services(args.config)
    else:
        SERVICES = app_api.Services()

    if args.run == 'OnBoot':
        on_boot()
    elif args.run == 'OnCommand':
        on_command()
    else:
        print("Unknown run level specified")
        sys.exit(1)

if __name__ == "__main__":
    main()
```

### DIFF
--- a/docs/getting-started/local-setup.rst
+++ b/docs/getting-started/local-setup.rst
@@ -34,7 +34,7 @@ After installing these dependencies, we suggest fetching the latest copy of the 
 
 After fetching the KubOS source repo, we suggest installing the Python ``app-api``, which is used in developing Python-based applications::
 
-    $ cd kubos/apis/app-api/python && pip3 install . --user
+    $ pushd kubos/apis/app-api/python && pip3 install . --user && popd
 
 After installing all of these dependencies, we suggest running the following script
 from the base of the kubos repo to verify everything is installed correctly and working::
--- a/docs/tutorials/first-mission-app.rst
+++ b/docs/tutorials/first-mission-app.rst
@@ -336,11 +336,9 @@ For each request, it:
     - Parses the response message and checks for errors
     - Returns the message payload if the request was successful
 
-To start, we'll import the API and create a constant for readability::
+To start, we'll import the API::
 
     import app_api
-    
-    SERVICES = app_api.services()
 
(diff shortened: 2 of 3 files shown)


docs/getting-started/local-setup.rst
```diff
@@ -34,7 +34,7 @@ After installing these dependencies, we suggest fetching the latest copy of the
 
 After fetching the KubOS source repo, we suggest installing the Python ``app-api``, which is used in developing Python-based applications::
 
-    $ cd kubos/apis/app-api/python && pip3 install . --user
+    $ pushd kubos/apis/app-api/python && pip3 install . --user && popd
```
plauche: Woah I didn't know about those commands!

PR #418, opened 2019-07-08
Tweaking first mission app to be compatible with OSX

CatGarab:
The monitor service relies on `/proc/meminfo` in order to query the system's available memory. OSX doesn't have this file, which means that the first mission app tutorial doesn't work for Mac users.

Tweaking the app to just do a `{ping}` query instead

### DIFF
--- a/docs/tutorials/app-register.rst
+++ b/docs/tutorials/app-register.rst
@@ -238,13 +238,13 @@ Updating
 --------
 
 After looking at our output, it would be nice if our memory message included the timestamp of
-when the system memory was checked.
+when the system was checked.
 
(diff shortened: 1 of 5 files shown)


CatGarab:
[Fix tutorials for OSX](https://trello.com/c/wby307J5/1016-fix-tutorials-for-osx)

marshall:
@CatGarab since this is a client-side only fix, doesn't this means that any request for `memInfo` with the app service running in an OS X environment will still yield an error of some kind?

CatGarab:
@marshall yes, that is correct. This is a documentation-only change

marshall:
@CatGarab k. I added a follow up to fix the underlying issue: https://trello.com/c/YCYhw2Bn/1018-implement-meminfo-for-macos

PR #428, opened 2019-07-11
Doc Review changes 

jacoffey3:
New section under mission needs for payload integration.

Linked real time section to design decisions.

### DIFF
--- a/docs/mission-dev/index.rst
+++ b/docs/mission-dev/index.rst
@@ -26,13 +26,11 @@ KubOS provides a payload service guide for integrating payloads into KubOS in th
 
 - :doc:`Payload Services <../ecosystem/services/payload-services>`
 
-Following this service guide will align it with how all other hardware services communicate in KubOS, but it is not *required*.
-To save development time, reduce complexity, or whatever reasoning you have, payloads can be integrated in whatever way your mission needs, as long as Linux supports it.
-Some examples:
-
-- Interact with your payload directly from a mission application using hardware interfaces (I2C, UART, etc.)
-- Create a UDP abstraction and access it directly from ground
-- Ethernet payload that accessible directly already using just the comm service
+We recommend all payload services follow this guide in order to keep system component interoperability simple and organized.
+However, there is no hard requirement forcing payloads to conform to our specifications.
+We want users to retain complete control over their systems and understand there may be cases where following our architecture isn't the best option.
+For example, there might be a payload device which requires a lower latency than our communication flow provides.
+In this case, it would be better for the appropriate mission application to be able to directly communicate with the device, rather than via a payload service.
 
 3. Mission Applications
 -----------------------
--- a/docs/mission-dev/index.rst
+++ b/docs/mission-dev/index.rst
@@ -20,6 +20,7 @@ To expose their downlink and uplink capability to the rest of the system, we pro
 
 2. Payload Integration(s)
 -------------------------
+
 Besides avionics hardware, you likely have a payload you want to integrate for your mission.
 KubOS provides a payload service guide for integrating payloads into KubOS in the *recommended* way:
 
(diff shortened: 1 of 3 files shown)


docs/mission-dev/index.rst
```diff
@@ -19,7 +18,22 @@ Radios have an additional step that's required to finish integrating them into t
 The hardware service only exposes the commands and telemetry for the radio hardware, but does not expose the uplink or downlink interface(s).
 To expose their downlink and uplink capability to the rest of the system, we provide a :doc:`communication service framework. <../ecosystem/services/comms-framework>`
 
-2. Mission Applications
+2. Payload Integration(s)
```
CatGarab: I had kind of assumed that this would be a sub-section under "Hardware Integrations", since payload services are a subset of hardware services

docs/mission-dev/index.rst
```diff
@@ -19,7 +18,22 @@ Radios have an additional step that's required to finish integrating them into t
...
+-------------------------
+Besides avionics hardware, you likely have a payload you want to integrate for your mission.
+KubOS provide a payload service framework for integrating payloads into KubOS in the *recommended* way:
```
CatGarab: - provide -> provides
- framework -> guide

docs/mission-dev/index.rst
```diff
@@ -19,7 +18,22 @@ Radios have an additional step that's required to finish integrating them into t
 The hardware service only exposes the commands and telemetry for the radio hardware, but does not expose the uplink or downlink interface(s).
 To expose their downlink and uplink capability to the rest of the system, we provide a :doc:`communication service framework. <../ecosystem/services/comms-framework>`
 
-2. Mission Applications
+2. Payload Integration(s)
+-------------------------
```
CatGarab: nit: Add an empty line after this

docs/mission-dev/index.rst
```diff
@@ -19,7 +18,22 @@ Radios have an additional step that's required to finish integrating them into t
...
+Besides avionics hardware, you likely have a payload you want to integrate for your mission.
+KubOS provide a payload service framework for integrating payloads into KubOS in the *recommended* way:
+
+- :doc:`Payload Services <../ecosystem/services/payload-services>`
+
+Following this service guide will align it with how all other hardware services communicate in KubOS, but it is not *required*.
```
CatGarab: This section is kind of awkward to me. Maybe this:

We recommend all payload services follow this guide in order to keep system component interoperability simple and organized. However, there is no hard requirement forcing services to conform to our specifications. We want users to retain complete control over their systems and understand there may be cases where following our architecture isn't the best option. For example, there might be a payload device which requires a lower latency than our communication flow provides. In this case, it would be better for the appropriate mission application to be able to directly communicate with the device, rather than going via a payload service.

...this ended up wordier than I normally like... >.>

docs/mission-dev/index.rst
jacoffey3: Not if we want to truly highlight it. Plus it's a much different step in the actual mission development cycle. I gave this some serious thought and I feel very strongly that it should be its own section, especially given the feedback we've gotten.

docs/mission-dev/index.rst
jacoffey3:
```suggestion
KubOS provides a payload service guide for integrating payloads into KubOS in the *recommended* way:
```

docs/mission-dev/index.rst
jacoffey3:
```suggestion
-------------------------

```

docs/mission-dev/index.rst
jacoffey3: I struggled with this too lol

docs/mission-dev/index.rst
```diff
@@ -10,7 +10,6 @@ The following steps don't have to be done precisely in this order, but this is g
 
 Although there is already quite a bit of :doc:`supported hardware <../index>` in KubOS, your mission might have some hardware that is not yet supported, or unique to your mission (such as a payload).
```
CatGarab: Maybe remove `(such as a payload)` since it's going to be explicitly called out in its own section

jacoffey3: Eh, reiterating is fine I think.

PR #429, opened 2019-07-12
Updating comms-service docs to reflect udp passthrough.

plauche:


plauche:
[Update comms service doc to reflect new udp passthrough method](https://trello.com/c/IRYZz7iS/932-update-comms-service-doc-to-reflect-new-udp-passthrough-method)

docs/ecosystem/services/comms-framework.rst
```diff
@@ -122,7 +128,7 @@ Once this transaction has completed, the message handler thread exits.
     box "Communications Service" #LightBlue
         participant "Read Thread" as read
 
-        Radio <- read : 1. Read data packets from radio
+        Radio -> read : 1. Read data packets from radio
         read -> read : 2. Deframe data packets
         read -> read : 3. Reassemble data packet
```
CatGarab: Maybe should have noticed this before now :P

Step 5 mentions GraphQL requests, while this section only refers to the more broad HTTP requests.

It would be good if this was commonized. Either mention GraphQL in the description above, or change Step 5 and 6 to refer to HTTP request/responses. I don't have particular feelings on which way you go.

jacoffey3: If our link layer supports HTTP and not just GraphQL, it should be HTTP. But I'm not sure if the comm service would support other types of HTTP

plauche: In the comms-service source we call it a GraphQL payload, not an HTTP payload, however there is nothing GraphQL specific about how the HTTP transaction is handled. I believe that any REST API endpoint which accepts a POST would work just fine with the current comms-service implementation.

PR #430, opened 2019-07-15
Moving crates with Rust & C out of the workspace's default-members

plauche:
Moving crates with mixed Rust & C source or dependencies out of the default build path for this Cargo workspace. 

This will allow for running commands like `cargo check` or `cargo test` at the root of the workspace without requiring all of the C build dependencies.

The whole expanded members of the workspace can be accessed with the `--all` flag, like `cargo check --all`.


plauche:
[Remove C-dependent services from top level cargo build so we can exclude C from the local building instructions (basically, if you're using C, use the SDK or figure it out yourself)](https://trello.com/c/snyKZ3Et/963-remove-c-dependent-services-from-top-level-cargo-build-so-we-can-exclude-c-from-the-local-building-instructions-basically-if-you)

PR #431, opened 2019-07-16
Making kubos_service::Config::new fail when the config can't be parsed

CatGarab:
Currently, if something fails while setting up a service, a default configuration is used, setting the service's URL to `127.0.0.1:8080`. Since nothing can tell that this default was used, any entity trying to interact with the service later on will fail. It is better for the issue to cause a failure as soon as possible, so it is more easily captured and understood.

This PR contains 3 main changes, and then a bunch of lines which had to be updated as a result:

- apis/system-api/src/config.rs
    - Updated all of the `Config::new*` functions to throw errors, rather than take the default values
    - Tweaked `hosturl()` to return `None` if no `addr` section was defined, rather than the default URL
- Updated all of the services to write error messages to the log if they fail, so that way we can see errors which occur when the system is booting


CatGarab:
[Update system API config parsing to not fail silently](https://trello.com/c/y1L9ov96/986-update-system-api-config-parsing-to-not-fail-silently)

jacoffey3:
[Combine vagrant boxes into single box](https://trello.com/c/YL7OMNgz/248-combine-vagrant-boxes-into-single-box)

PR #436, opened 2019-07-24
Updating all Python package requirements.txt files

CatGarab:
The intention is that you can install all dependencies by running `pip3 install -r requirements.txt` from the repo's root directory

Most packages are version locked based on the version being used by Buildroot

### DIFF
--- a/apis/app-api/python/requirements.txt
+++ b/apis/app-api/python/requirements.txt
@@ -1,0 +1,5 @@
+mock==3.0.5
+requests==2.21.0
+responses==0.10.6
+setuptools==40.6.3
+toml==0.9.3
--- a/apis/app-api/python/requirements.txt
+++ b/apis/app-api/python/requirements.txt
@@ -2,4 +2,4 @@ mock==3.0.5
 requests==2.21.0
 responses==0.10.6
 setuptools==40.6.3
-toml==0.9.3
+toml==0.9.3.1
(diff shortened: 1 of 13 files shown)


CatGarab:
[Top level requirements.txt for all the python packages (maybe one that just points at the other requirements.txt files?)](https://trello.com/c/b83ZVCVI/976-top-level-requirementstxt-for-all-the-python-packages-maybe-one-that-just-points-at-the-other-requirementstxt-files)

requirements.txt
```diff
@@ -0,0 +1,10 @@
+-r apis/app-api/python/requirements.txt
+-r apis/pumpkin-mcu-api/requirements.txt
+-r examples/pumpkin-mcu-api-example/requirements.txt
+-r examples/python-mission-application/requirements.txt
+-r examples/python-mission-framework/requirements.txt
+-r examples/python-service/requirements.txt
+-r hal/python-hal/i2c/requirements.txt
+-r libs/kubos-service/kubos_service/requirements.txt
```
plauche: This requirements file is missing.

apis/app-api/python/requirements.txt
```diff
@@ -0,0 +1,5 @@
+mock==3.0.5
+requests==2.21.0
+responses==0.10.6
+setuptools==40.6.3
+toml==0.9.3
```
plauche: I had to move the `toml` version to `0.9.3.1`, it that `0.9.3` does not exist.

CatGarab: Interestingly, Buildroot is still pulling `0.9.3` successfully. So I guess we'll just have a slight version mismatch there

PR #438, opened 2019-07-24
Adding Rust compile-size tips

CatGarab:


CatGarab:
[Add Rust binary size optimization tips to FAQ](https://trello.com/c/s5bqjtnQ/961-add-rust-binary-size-optimization-tips-to-faq)

jacoffey3:
I don't think you should repeat it 3 times. It would be better to pick a spot and link to it from the others.

PR #439, opened 2019-07-24
App Service waits 300ms to ensure it catches Python crashes

cantino:


jacoffey3:
[Fix test_python_app on Mac and look at why app service isn't failing](https://trello.com/c/r2Q6Z6YW/1033-fix-testpythonapp-on-mac-and-look-at-why-app-service-isnt-failing)

PR #443, opened 2019-08-01
MVP Scheduler Service

### LINKED ISSUE #336: Adding a Scheduler  to KubOS
We want to add native schedule support to the KubOS system for Mission Applications to leverage. Currently, mission apps have to manage timed tasks internally for both recurring and one-time tasks. We want to move that logic to the KubOS system as a whole, to keep Mission Applications simpler. 

Assumptions
-------------
- System time is our source of truth

Requirements
--------------
- Schedule task for execution at a future time/date
  - Second level granularity for MVP
  - Millisecond (depending on resources) for ideal 
- Schedule task for recurring execution
  - Every X time interval
  - Start time
  - End time
  - Total executions
- Preserve schedule of tasks across reboots
- Tasks can consist of
  - Mission application service commands
  - Any executable available on the file system
  - Any command which can be executed via /bin/sh
- Commands
  - Add task to schedule
  - Remove task from schedule
- Telemetry
  - List of scheduled tasks and their schedules
- Scheduler control
  - Pause/Resume
- Should it have the ability to run tasks if they are missed due to system downtime?

plauche:
Closes #336 
- [x] Service Skeleton #444 
- [x] Basic mutations implementation #446 
- [x] Basic queries implementation #450 
- [x] Init tasks #465 
- [x] Scheduler Modes #477 
- [x] One-time tasks #479 
- [x] Recurring tasks #479 
- [x] Safe mode & failover #479 
- [x] Raw Task Mutation #483


plauche:
[#Issue 336 ⁃ Adding a Scheduler  to KubOS](https://trello.com/c/heCXiO5B/995-issue-336-adding-a-scheduler-to-kubos)

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,153 @@
...
+The ``switchSchedule`` mutation allows the scheduler to switch between two different schedule modes.
+It has the following schema::
+
+    mutation {
+        switchSchedule(name: String!): {
+            success: Boolean!
+        }
+    }
+
+The ``registerSchedule`` mutation allows the scheduler to register a new schedule file. It has
```
CatGarab: nit: You're not going to be registering anything else, so this mutation could just be `register`

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,153 @@
...
+Tasks in the ``recurring`` section will be executed on a recurring basis. A ``start-time`` and
+``end-time`` are given to supply bounds around the occurance of the task. The task
+will occur at the given frequency beginning at the ``start-time``. Each task
+in this section will be specified like so:
+
+.. code-block:: json
+
+    {
+        "task-name": {
+            "start-time": "start time of execution in yyyy-mm-dd-hh:mm:ss format",
```
CatGarab: nit: date/time format should follow [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Combined_date_and_time_representations), so the delimiter between `dd` and `hh` should be either a `T` or a space

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,153 @@
...
+            "end-time": "end time of execution in yyyy-mm-dd-hh:mm:ss format",
+            "frequency": "frequency of execution in HH:mm:ss format",
```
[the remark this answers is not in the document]
plauche: Would it be better to add an extra digit to the hours field or to add a days field?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,153 @@
...
+            "command": "task command"
```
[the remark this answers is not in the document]
plauche: I looked back through our cards and we decided to only implement app execution for MVP, so that cleans this up :sunglasses:.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,153 @@
...
+        },
+        "recurring": {
+            "clean-logs-every-12hrs": {
+                "start-time": "2019-08-11-15:20:10"
+                "end-time": "2019-08-12-15:20:10"
+                "frequency": "12:00:00",
+                "command": "rm -rf /tmp/logs/*"
+            }
+        }
+    }
```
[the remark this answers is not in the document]
plauche: Yes that would be great.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,153 @@
...
+Mutations
+~~~~~~~~~
+
+The scheduler has two mutations: ``switchSchedule`` and ``registerSchedule``. 
+
+The ``switchSchedule`` mutation allows the scheduler to switch between two different schedule modes.
+It has the following schema::
+
+    mutation {
+        switchSchedule(name: String!): {
```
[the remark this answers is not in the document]
plauche: :+1: for name

docs/ecosystem/services/scheduler.rst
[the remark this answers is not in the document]
CatGarab: days field would probably be better. It's more human-readable

cantino:
Are we moving to schedules being directories?

jacoffey3:
> Are we moving to schedules being directories?

No, not for MVP. We'll revisit that after talking to AFIT and doing more research/characterization.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+
+The KubOS system includes a scheduler service to facilitate recurring tasks
+and one time tasks with specific timing requirements.
+
+Behavior
+--------
+
+Upon boot, or service start, the scheduler will read the active schedule file and
+load the schedule into memory. The default active schedule file will be found at
+``/home/system/etc/schedules/active.json``. Any ``init`` tasks will be performed
```
jacoffey3: What order will the init tasks be completed in?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+immediately. All other ``one-time`` or ``recurring`` tasks will be scheduled
```
jacoffey3: Will recurring start immediately, or after their first period? Can I specify a delay time after schedule entry to start recurring?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+Behavior
+--------
+
+Upon boot, or service start, the scheduler will read the active schedule file and
+load the schedule into memory. The default active schedule file will be found at
+``/home/system/etc/schedules/active.json``. Any ``init`` tasks will be performed
+immediately. All other ``one-time`` or ``recurring`` tasks will be scheduled 
+to run at their designated times.
+
+By default the scheduler will have two schedules: ``operational`` and ``safemode``. These
```
jacoffey3: I'm actually onboard with @CatGarab with the modes vs names. I think having many named schedules is fine. I realized what we need more than "modes" generally, is the ability to specify a "safemode" schedule that is used as a fallback if the active schedule can't be read, and can be stored redundantly.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+schedules are represented by different schedule files maintained in the schedules directory.
+Only one schedule can be active at any given time.
+
+The scheduler will have its own log file, ``/var/log/kubos-schedule.log``, which
```
jacoffey3: Now that I think about it - should all core services have their own logs?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+----------------------
+
+Schedules will be specified in files in the `json` format. Each schedule file will contain
+all of the necessary information for that specific schedule. Different schedules, such as
+operational, safe mode, etc, will each have their own schedule files.
+
+Schedules consist of three sections: ``init``, ``one-time``, and ``recurring``. Each section
+represents a different type of scheduled task.
+
+Tasks in the ``init`` section will be executed on boot or on schedule change. Each
```
jacoffey3: Maybe here is a better place to describe how init tasks will be executed. Adding a init delay will be necessary I imagine, but there is a possibility that it's too complicated for MVP

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+represents a different type of scheduled task.
+
+Tasks in the ``init`` section will be executed on boot or on schedule change. Each 
+task in this section will be specified like so:
+
+.. code-block:: json
+
+    {
+        "task-name": {
+            "app-name": "task app",
```
jacoffey3: preceeding everything with "app" is probably not necessary. Maybe make this: 
```
{
  "app": {
    "name": "task app",
    "args": "args",
    "config": "path to config"
  }
  schedule fields...
}
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+Tasks in the ``one-time`` section will be executed once at a set time. Each task
+in this section will be specified like so:
+
+.. code-block:: json
+
+    {
+        "task-name": {
+            "time": "time of execution in yyyy-mm-dd hh:mm:ss format",
```
jacoffey3: For instance, like I mentioned before: 
```
{
  "app": {
    "name": "task app",
    "args": "args",
    "config": "path to config"
  },
  "time": "yyyy-mm-dd hh:mm:ss"
}
```

This way, we can add cmd line commands later with: 

```
{
  "cmd": "executable to run with args",
  "time": "yyyy-mm-dd hh:mm:ss"
}
```

and we will just have to detect if `app` or `cmd` are present.

docs/ecosystem/services/scheduler.rst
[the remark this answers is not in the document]
jacoffey3: It should also check to see if the apps you're telling it to start are present if possible

docs/ecosystem/services/scheduler.rst
cantino: Does this use syslog?

docs/ecosystem/services/scheduler.rst
cantino: I think init delay will be needed to allow for relative time command sequences.

docs/ecosystem/services/scheduler.rst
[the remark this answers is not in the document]
cantino: To get format ideas from a very full-featured scheduler, skim the readme for https://github.com/jmettraux/rufus-scheduler

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+
+    {
+        "init": {
+            "start-camera": {
+                "app-name": "activate-camera"
+            }
+        },
+        "one-time": {
+            "deploy-solar": {
+                "time": "2019-08-11 15:20:10"
```
cantino: Maybe this is a silly question, but what happens if the clock on the satellite is wrong?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+    {
+        "task-name": {
+            "time": "time of execution in yyyy-mm-dd hh:mm:ss format",
+            "app-name": "task app",
+            "app-args": "optional app args",
+            "app-config": "optional path to app config"
+        }
+    }
+
+Tasks in the ``recurring`` section will be executed on a recurring basis. A ``start-time`` and
```
cantino: I'd suggest that start and end should be optional, in which case it just goes forever. Also makes the system work even if the clock is wrong.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,159 @@
...
+will log all schedule related actions the scheduler takes.
+
+Schedule Specification
+----------------------
+
+Schedules will be specified in files in the `json` format. Each schedule file will contain
+all of the necessary information for that specific schedule. Different schedules, such as
+operational, safe mode, etc, will each have their own schedule files.
+
+Schedules consist of three sections: ``init``, ``one-time``, and ``recurring``. Each section
```
cantino: Idea: should there be any `unload` items to run when leaving a schedule?

docs/ecosystem/services/scheduler.rst
jacoffey3: Then it doesn't run it. Our core assumption is that system time is our source of truth. We are proceeding as if it's right. We need to account for if it's wrong in other ways.

docs/ecosystem/services/scheduler.rst
plauche: I was thinking the order that they are listed in the schedule file, but also open to other ideas.

docs/ecosystem/services/scheduler.rst
plauche: Because we are specifying a `startTime` and `frequency` for recurring tasks, I was thinking that we would schedule based on those two items. Recurring tasks won't start immediately, but at the end of the first period which hasn't already occurred.

docs/ecosystem/services/scheduler.rst
plauche: Currently we specify that all services and apps use `syslog` to collect logging. Individual logs for each service would make them more human readable without the need for grepping and parsing. It would also make it quicker to downlink logs specific to one part of the system. 

However if we instrument our logs with the correct information, then it would be simple to parse out the specific service logs from a singular collective log file. The one collective log file  also makes it a bit easier to see the whole state of the system around a certain time, which could be useful for diagnostics. On the other hand, if the logs are meant to be digested, reconstructed, and viewed through a tool like MT, then it does not really matter how we package them onboard.

docs/ecosystem/services/scheduler.rst
plauche: Yup I agree. Originally I was writing with app and non-app execution in mind, but if we are always running apps then prepending with `app` is not needed.

docs/ecosystem/services/scheduler.rst
plauche: This seems pretty reasonable.

docs/ecosystem/services/scheduler.rst
plauche: Hmmm good question. @jacoffey3 is this a use case people expect?

docs/ecosystem/services/scheduler.rst
plauche: That is fine. If the `start-time` is not included, then the recurring intervals will begin when the scheduler is started?

docs/ecosystem/services/scheduler.rst
plauche: Hmm maybe we should consider sequence related functionality in a post-MVP PR and have a discussion about different ways to accomplish that.

docs/ecosystem/services/scheduler.rst
plauche: This makes sense. Should we allow the specification of a safemode directory of schedules, or just a singular schedule file?

docs/ecosystem/services/scheduler.rst
jacoffey3: So I'd thought about this, and I don't think it's necessary for MVP. I agree that it's something that people might want, but I think the use case can also be satisfied by tackling those items when entering a new schedule. 

Generally, I assume people will kill all running apps when entering safemode to start from a zero state. The only thing that's not tackled with that is if they run another long running executable on init. But, they should put logic in to kill that for their safemode. I'd rather depend on my safemode schedule forcing the system into the appropriate state than trusting others to do proper unloading.

docs/ecosystem/services/scheduler.rst
jacoffey3: Order in the schedule file is good, just make sure that's documented.

docs/ecosystem/services/scheduler.rst
jacoffey3: I think for recurring, a relative delay time might actually be more appropriate than a start and end. I don't think start and end will almost ever get used

docs/ecosystem/services/scheduler.rst
jacoffey3: Are we supporting a directory currently?

docs/ecosystem/services/scheduler.rst
jacoffey3: This is another point where having a delay would be best. After X time start recurring task.

docs/ecosystem/services/scheduler.rst
jacoffey3: @plauche I think you're right that we should just dump the scheduler logs in with the rest of the services. If we need them parsed out, we can do so after the fact.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+when their corresponding delay has passed. All other tasks configured with a
+``time`` or ``period`` will be scheduled to run at their designated times.
+
+By default the scheduler comes with a single mode: ``safe``. This mode is reserved as a
+fallback and default operating mode. Additional modes can be created and made active
+through the GraphQL interface. All modes are represented as different directories in
+the schedules directory (``/home/system/etc/schedules``).
+Only one mode can be active at any given time.
+
+The scheduler has its own log file, ``/var/log/kubos-schedule.log``, which
```
jacoffey3: This is not true any more, correct?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+The KubOS system includes a scheduler service to facilitate recurring
+and one time tasks with specific timing requirements.
+
+Behavior
+--------
+
+The behavior of the scheduler is defined by three levels of organization: *mode*,
+*schedule*, and *task*. The *mode* determines which schedule directory is being executed.
+The *schedule* is then assembled by combining all the *task lists* in the schedule
+directory, each of which can contain one or many *tasks*. New *tasks* can be added to
```
jacoffey3: Make this a new paragraph to break up the wall of text. 
```suggestion
directory, each of which can contain one or many *tasks*. 

Adding Tasks
____________

New *tasks* can be added to
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+Schedule Specification
+----------------------
+
+Schedules are made up of tasks, which are specified through task lists in the 
+`json` format. Each task list contains all of the necessary information for each
+task to be scheduled. Multiple task lists can coexist in the same mode folder,
+allowing for easy loading of new scheduled tasks.
+
+Tasks can have their scheduled time of execution specified using three different
+fields: ``delay``, ``time``, and ``period``. The ``delay`` field specifies
```
jacoffey3: Are any of these mutually exclusive?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
+Scheduler
+=========
+
+The KubOS system includes a scheduler service to facilitate recurring
+and one time tasks with specific timing requirements.
+
+Behavior
+--------
+
+The behavior of the scheduler is defined by three levels of organization: *mode*,
```
jacoffey3:
```suggestion
Terminology
___________

The behavior of the scheduler is defined by three levels of organization: *mode*,
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+*schedule*, and *task*. The *mode* determines which schedule directory is being executed.
+The *schedule* is then assembled by combining all the *task lists* in the schedule
+directory, each of which can contain one or many *tasks*. New *tasks* can be added to
+a *schedule* by uploading a *task list* to a location which is accessible to the 
+scheduler, and importing that *task list* to the appropriate mode. *Tasks* can
+also be added directly via the ``importRawTaskList`` GraphQL mutation.
+
+Upon boot, or service start, the scheduler reads all task lists in the active
```
jacoffey3:
```suggestion
Bootup Behavior
_______________

Upon boot, or service start, the scheduler reads all task lists in the active 
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+a *schedule* by uploading a *task list* to a location which is accessible to the 
+scheduler, and importing that *task list* to the appropriate mode. *Tasks* can
+also be added directly via the ``importRawTaskList`` GraphQL mutation.
+
+Upon boot, or service start, the scheduler reads all task lists in the active 
+mode directory and schedules all tasks from that mode. The default active mode directory
+is found at ``/home/system/etc/schedules/active``, which is a symlink
+to the actual active mode's directory.
+
+Schedules consist of many tasks, each of which has a scheduled execution time.
```
jacoffey3: This is a redundant section. It should be in the Task description in the next section

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+
+By default the scheduler comes with a single mode: ``safe``. This mode is reserved as a
+fallback and default operating mode. Additional modes can be created and made active
+through the GraphQL interface. All modes are represented as different directories in
+the schedules directory (``/home/system/etc/schedules``).
+Only one mode can be active at any given time.
+
+The scheduler has its own log file, ``/var/log/kubos-schedule.log``, which
+logs all schedule related actions the scheduler takes.
+
```
jacoffey3: Add a new section describing when we go into ``safe`` mode automatically, and title it accordingly

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+to the actual active mode's directory.
+
+Schedules consist of many tasks, each of which has a scheduled execution time.
+Tasks are loaded into the scheduler when the scheduler service starts,
+when a mode becomes active, or when a new task list is imported into the active mode.
+Any tasks configured with a ``delay`` will be scheduled for execution
+when their corresponding delay has passed. All other tasks configured with a
+``time`` or ``period`` will be scheduled to run at their designated times.
+
+By default the scheduler comes with a single mode: ``safe``. This mode is reserved as a
```
jacoffey3:
```suggestion
Scheduler Modes
_______________

By default the scheduler comes with a single mode: ``safe``. This mode is reserved as a
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
+Scheduler
+=========
+
+The KubOS system includes a scheduler service to facilitate recurring
+and one time tasks with specific timing requirements.
+
+Behavior
```
jacoffey3: This is a really huge wall of text, I'm suggesting some subsections for it rather than just paragraphs.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
+Scheduler
+=========
+
+The KubOS system includes a scheduler service to facilitate recurring
+and one time tasks with specific timing requirements.
```
jacoffey3:
```suggestion
and one time tasks with specific timing requirements, 
and to facilitate managing these tasks in groups of system-level modes. 
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,332 @@
...
+through the GraphQL interface. All modes are represented as different directories in
+the schedules directory (``/home/system/etc/schedules``).
+Only one mode can be active at any given time.
+
+The scheduler has its own log file, ``/var/log/kubos-schedule.log``, which
+logs all schedule related actions the scheduler takes.
+
+.. _schedule-specification:
+
+Schedule Specification
```
jacoffey3: Maybe this section would be better titled "How to build a Task List" 

It honestly looks a lot like things that would fall under "Behavior" as well, as you're describing how `delay`, `period`, and `time` work, and what Tasks are able to do.

docs/ecosystem/services/scheduler.rst
plauche: Correct!

docs/ecosystem/services/scheduler.rst
plauche: Yes, now that I look over it again there is a fair amount of format specification and behaviour mixed together. Maybe we need separate `Service Behaviour` and `Task Behaviour` sections for the higher level service operations and the more fine grained scheduling behaviour?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,351 @@
...
+        "delay": "Required start delay in Xh Ym Zs format",
+        "period": "Required period of execution in Xh Ym Zs format",
+        "app": {
+            "name": "Required registered name of app to run",
+            "args": ["Optional", "command", "line", "app", "args"],
+            "config": "Optional path to app config"
+        }
+    }
+
+An example schedule config:
```
jacoffey3:
```suggestion
An example task list:
```

docs/tutorials/schedule-app.rst
```diff
@@ -0,0 +1,195 @@
...
+
+This tutorial will use the following example directories:
+
+    - ``/home/user/kubos`` - Cloned copy of the kubos responsible
+    - ``/home/user/kubos/schedules`` - Directory used by the schedules service
+      to store schedule information
+
+- Navigate to ``/home/user/kubos`` (or your preferred copy of the kubos repo).
+
+- Run the following command to start the scheduler service in the background
```
jacoffey3: Should you run it like that? Or by doing the `cargo run --bin scheduler-service -- -c tools/local_config.toml` command? Also should they be starting it in the background? If they do they won't be able to kill it.

docs/ecosystem/services/scheduler.rst
jacoffey3: Yeah that makes sense! Maybe just a "Tasks and How to Make Them" Section? Tells you what a task is, how it behaves, and how to build a Task List

docs/tutorials/schedule-app.rst
jacoffey3: Nevermind, it's like that in all tutorials. We'll need to change that at some point.

docs/tutorials/schedule-app.rst
```diff
@@ -0,0 +1,195 @@
...
+                "errors": ""
+            }
+        }
+    }
+
+At this point our mode and task list have been loaded into the scheduler and will begin execution.
+The console where you started the app service should show the app's execution messages after 10
+seconds and then every 10 seconds after that::
+
+    Current available memory: 4390792 kB
```
jacoffey3: This is what the tutorial app outputs
```suggestion
    Successfully pinged monitor service
```

jacoffey3:
@plauche Tutorial worked great! Starting to test general behavior.

jacoffey3:
@plauche What does every level of the mode have a name? It seems like the mode should have a name, and the task list should probably be named "filename" or something like that. It looks like each task actually can repeat the same name in the same file, which means it's more of a description than a name. You could change that to be description and make it optional? 

```
query {availableModes{
  name
  path
  lastRevised
  active
  schedule {
    name
    path
    timeImported
    tasks {
      name
      delay
      time
      period
      app {
        name
        args
        config
      }
    }
  }
}}```

**Comment by jacoffey3** [comment]
Needs validation of the entire file on import.

File: docs/ecosystem/services/scheduler.rst [hunk 31]
```diff
@@ -0,0 +1,397 @@
...
+when the task will be executed. The ``period`` field indicates the app should
+be executed on a recurring basis and specifies the period of recurrence. The ``delay``
+field is required, except when using the ``time`` field. The ``time`` and ``period``
+fields may not be used together.
+
+Delayed Tasks
+~~~~~~~~~~~~~
+
+Tasks configured with only a ``delay`` will be executed on boot or on schedule change.
+Tasks will be assigned to the scheduler in an unpredictable order. The actual execution time
```
jacoffey3: You restate this in the next line. 
```suggestion
The actual execution time
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+            config: String,
+        }
+    }
+
+
+Mutations
+~~~~~~~~~
+
+The scheduler also exposes the following mutations: ``createMode``, ``removeMode``,
+``activateMode``, ``importTaskList``, and ``removeTaskList``.
```
jacoffey3:
```suggestion
``activateMode``, ``importTaskList``, ``removeTaskList``, and ``safeMode``.
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+            errors: String
+        }
+    }
+
+Activating Modes
+~~~~~~~~~~~~~~~~
+
+The ``activateMode`` mutation instructs the scheduler to make the specified mode
+active. It cannot be used to activate safe mode, the ``safeMode`` mutation is the
+only way to activate safe mode. It has the following schema::
```
jacoffey3: Indefinite "it"
```suggestion
only way to activate safe mode.``activeMode`` has the following schema::
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+
+Activating Safe Mode
+~~~~~~~~~~~~~~~~~~~~
+
+The ``safeMode`` mutation instructs the scheduler to make the *safe* mode
+active. This mutation is the only way to activate *safe* mode and can only
+activate that mode. It has the following schema::
+
+    mutation {
+        safeMode(name: String!): {
```
jacoffey3: Does it need the `name` arg? Can we remove it? 
```suggestion
        safeMode(): {
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+            errors: String
+        }
+    }
+
+Importing Task Lists
+~~~~~~~~~~~~~~~~~~~~
+
+The ``importTaskList`` mutation allows the scheduler to import a new task list into
+a specified mode. If the targeted mode is active, all tasks in the task list will be
+immediately loaded for scheduling. It has the following schema::
```
jacoffey3:
```suggestion
immediately scheduled. It has the following schema::
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+            errors: String
+        }
+    }
+
+Removing Task Lists
+~~~~~~~~~~~~~~~~~~~
+
+The ``removeTaskList`` mutation allows the scheduler to remove a task list from
+a specified mode. If the mode is active, all tasks in the task list will be removed
+from the scheduler. It as the following schema::
```
jacoffey3:
```suggestion
from the schedule. It as the following schema::
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+=========
+
+The KubOS system includes a scheduler service to facilitate recurring
+and one time tasks with specific timing requirements, and to facilitate
+managing these tasks in groups of system-level modes. 
+
+Behavior
+--------
+
+Terminology
```
jacoffey3: You should add a diagram to illustrate this.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+*schedule*, and *task*. The *mode* determines which schedule directory is being executed.
+The *schedule* is then assembled by combining all the *task lists* in the schedule
+directory, each of which can contain one or many *tasks*. 
+
+Adding Tasks
+~~~~~~~~~~~~
+
+New *tasks* can be added to a *schedule* by uploading a *task list* to a location
+which is accessible to the scheduler, and importing that *task list* to the appropriate
+mode. *Tasks* can also be added directly via the ``importRawTaskList`` GraphQL mutation.
```
jacoffey3: Overly precise for this section of the doc. I don't want them going to look it up.
```suggestion
mode. *Tasks* can also be added directly via GraphQL interface. 
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,397 @@
...
+Failover Behavior
+~~~~~~~~~~~~~~~~~
+
+In addition to using the ``safe`` mode as the default operating mode, the scheduler is 
+designed to automatically failover to the ``safe`` mode when any problems are detected.
+The following situations will cause the scheduler to activate the ``safe`` mode::
+
+    - The scheduler starts and no modes exist
+    - The scheduler starts and no mode is active
+    - The scheduler attempts to activate a mode and the mode does not exist
```
jacoffey3: Did we talk about if the scheduler is unable to parse a task list? Or if the task list fails checksum?

docs/ecosystem/services/scheduler.rst
plauche: Ahh that's a copy/paste error, that arg doesn't actually exist.

docs/ecosystem/services/scheduler.rst
plauche: We have not taken that into consideration with the fail over coverage. We also don't take into consideration the situation where apps in the current mode fail to start and return errors through the `startApp` mutation.

docs/ecosystem/services/scheduler.rst
cantino: via *the* GraphQL interface.

cantino:
Something weird happened and the `scheduler.rst` file I was reviewing disappeared.

docs/ecosystem/services/scheduler.rst
jacoffey3: What happens? That behavior should be documented.

services/scheduler-service/src/app.rs
```diff
@@ -0,0 +1,104 @@
...
+
+impl App {
+    pub fn execute(&self, service_url: &str) {
+        info!("Start app {}", self.name);
+        let mut query_args = format!("name: \"{}\"", self.name);
+        if let Some(config) = &self.config {
+            query_args.push_str(&format!(", config: \"{}\"", config));
+        }
+        if let Some(args) = &self.args {
+            let app_args: Vec<String> = args.iter().map(|x| format!("\"{}\"", x)).collect();
```
cantino: Is `format!("\"{}\"", x)` being used to quote the arg? What if it has quotes in it?

services/scheduler-service/src/app.rs
```diff
@@ -0,0 +1,104 @@
...
+// Configuration used for execution of an app
+#[derive(Clone, Debug, GraphQLObject, Serialize, Deserialize)]
+pub struct App {
+    pub name: String,
+    pub args: Option<Vec<String>>,
+    pub config: Option<String>,
+}
+
+impl App {
+    pub fn execute(&self, service_url: &str) {
```
cantino: Should this return a `Result` so that we can go into safe mode?

services/scheduler-service/src/mode.rs
```diff
@@ -0,0 +1,240 @@
...
+    let name = name.to_lowercase();
+    if let Ok(Some(active_mode)) = get_active_mode(scheduler_dir) {
+        name == active_mode.name
+    } else {
+        false
+    }
+}
+
+pub fn get_available_modes(
+    scheduler_dir: &str,
```
cantino: Are dirs always assumed to be absolute?

services/scheduler-service/src/scheduler.rs
```diff
@@ -0,0 +1,196 @@
...
+    // Iterate through the active mode and kick off scheduling tasks
+    pub fn start(&self) -> Result<(), SchedulerError> {
+        if let Some(active_mode) = get_active_mode(&self.scheduler_dir)? {
+            for list in get_mode_task_lists(&active_mode.path)? {
+                self.start_task_list(list)?;
+            }
+            Ok(())
+        } else {
+            error!("Failed to find an active mode");
+            panic!("Failed to find an active mode");
```
cantino: Should this switch to safe mode instead of panicking?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,416 @@
...
+~~~~~~~~~~~~~~
+
+Tasks configured with a ``time`` field will be executed once at a set time. Each 
+one time task is specified like so:
+
+.. code-block:: json
+
+    {
+        "description": "Task description",
+        "time": "Required time of execution in yyyy-mm-dd hh:mm:ss format",
```
cantino: Will this always be UTC?

services/scheduler-service/src/task_list.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// Copy a task list into a mode directory
+pub fn import_task_list(
+    scheduler_dir: &str,
+    raw_name: &str,
+    path: &str,
+    raw_mode: &str,
+) -> Result<(), SchedulerError> {
+    let name = raw_name.to_lowercase();
```
cantino: Do we document that modes / names should be lowercase?

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,416 @@
...
+            "args": ["Optional", "command", "line", "app", "args"],
+            "config": "Optional path to app config"
+        }
+    }
+
+Recurring Tasks
+~~~~~~~~~~~~~~~
+
+Tasks configured with a ``period`` field will be executed on a recurring basis. The task
+will occur at the given ``period`` beginning after the ``delay`` has expired.
```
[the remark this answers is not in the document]
plauche: Ah yes, that reads much better.

services/scheduler-service/src/app.rs
plauche: Yes, it is being used to create the GraphQL query, which requires that the app args get placed within quotes, ie `startApp(args: "-l arg") {}`. The args are quoted strings in the JSON format, so any quotes will already be escaped, or the JSON won't parse correctly.

services/scheduler-service/src/app.rs
plauche: We don't currently plan on failing over when apps fail to execute.

services/scheduler-service/src/mode.rs
plauche: The `scheduler_dir` is assumed to always be absolute because we convert it to an absolute dir after reading it from the config file.

services/scheduler-service/src/scheduler.rs
plauche: Currently `start` is called in three scenarios:
- The scheduler has just started and has just verified an active mode exists, or failed over to safe mode.
- The scheduler has just successfully switched to a new active mode, which includes the possibility of failing over on certain errors.
- The scheduler has just switched to safe mode. 

I think it is a reasonable assumption that if we can't find an active mode immediately after any of those three scenarios, we are in a bad place and need to panic out.

plauche:
@cantino when a task list is added we validate the json, we enforce the required fields, and we validate the various time field.s

services/scheduler-service/src/app.rs
jacoffey3: @plauche it should log if they fail though.

docs/ecosystem/services/scheduler.rst
cantino: If so, document expectation of UTC everywhere.

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,434 @@
...
+    - The scheduler starts and no modes exist
+    - The scheduler starts and no mode is active
+    - The scheduler attempts to activate a mode and the mode does not exist
+    - The scheduler loads an active mode and encounters an invalid task list
+
+The following situations will **not** cause the scheduler to activate ``safe`` mode:
+
+    - The scheduler loads an active mode and encounters a task scheduled for the past
+    - The scheduler encounters errors when querying the application service to start an app
+    - The scheduler imports a task list into an active mode and is unable to parse or validate the list
```
jacoffey3: Not worth mentioning here, as we mention that we validate tasks on import. 
```suggestion
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,434 @@
...
+In addition to using the ``safe`` mode as the default operating mode, the scheduler is 
+designed to automatically failover to the ``safe`` mode when any problems are detected.
+The following situations will cause the scheduler to activate the ``safe`` mode:
+
+    - The scheduler starts and no modes exist
+    - The scheduler starts and no mode is active
+    - The scheduler attempts to activate a mode and the mode does not exist
+    - The scheduler loads an active mode and encounters an invalid task list
+
+The following situations will **not** cause the scheduler to activate ``safe`` mode:
```
jacoffey3: I guess it would have been better to say we need to highlight the fact that we aren't attempting to recover automatically from an app not starting properly. That's all that needs highlighting. 

```suggestion
The only error situation that will **not** cause the scheduler to activate ``safe`` mode is if the scheduler encounters errors when querying the application service to start an app. 
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,434 @@
...
+
+    - The scheduler loads an active mode and encounters a task scheduled for the past
```
jacoffey3: Not worth including, as it should be obvious. 
```suggestion
```

docs/ecosystem/services/scheduler.rst
```diff
@@ -0,0 +1,434 @@
...
+    - The scheduler encounters errors when querying the application service to start an app
```
jacoffey3: Removing due to the wording change I suggested above. 
```suggestion
```

PR #445, opened 2019-08-06
Adding app monitoring capabilities

CatGarab:


CatGarab:
[Add process monitoring to applications started to the mission application service](https://trello.com/c/PkMa5HyS/783-add-process-monitoring-to-applications-started-to-the-mission-application-service)

CatGarab:
[Add process monitoring to applications started to the mission application service](https://trello.com/c/PkMa5HyS/783-add-process-monitoring-to-applications-started-to-the-mission-application-service)

services/app-service/src/registry.rs
```diff
@@ -611,17 +622,75 @@ impl AppRegistry {
...
             }
-        } else {
-            Ok(child.id())
+            Ok(None) => {
+                let name = app_name.to_owned();
+                let pid = child.id() as i32;
+                let run_level_str = format!("{}", run_level);
+                let registry = self.monitoring.clone();
+
+                // Check if app is already running
```
plauche: Should we check if the app is already running prior to `cmd.spawn`?

services/app-service/src/schema.rs
```diff
@@ -66,6 +67,28 @@ graphql_object!(QueryRoot : Context as "Query" |&self| {
...
         Ok(result)
     }
+
+    field running_apps(&executor,
+        name: Option<String>)
+        -> FieldResult<Vec<MonitorEntry>> as "Running Apps Query"
+    {
+        let mut result: Vec<MonitorEntry> = Vec::new();
+        let entries = executor.context().subsystem().monitoring.lock()?;
+        let final_iter = entries.iter().filter(|e| {
```
plauche: Could you compact this into:
```
        let result = entries.iter().filter(|e| {
            if name.is_some() && &e.name != name.as_ref().unwrap() {
                return false;
            }
            true
        }).map(|entry| (*entry).clone()).collect();
```

services/app-service/tests/test_monitor.rs
```diff
@@ -0,0 +1,250 @@
...
+    send_query(
+        config.clone(),
+        r#"mutation {
+            startApp(name: "rust-proj", runLevel: "OnCommand", args: "-l") {
+                errors
+            }
+        }"#,
+    );
+
+    // If we try to start the app a second time, it should fail
```
plauche: I think this comment might be incorrect?

services/app-service/src/registry.rs
CatGarab: ....yes. yes we should.

docs/tutorials/app-register.rst
```diff
@@ -213,7 +213,8 @@ The mutation returns three fields:
 
     - ``success`` - Indicating the overall result of the operation
     - ``errors`` - Any errors which were encountered while starting the application
-    - ``pid`` - The PID of the started application. This will be empty if any errors are encountered
+    - ``pid`` - The PID of the started application. This will be empty if any errors are encountered,
```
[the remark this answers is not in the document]
CatGarab: If the app has already finished, the PID is no longer useful. The system frees it up so it can be used for another process

docs/ecosystem/services/app-service.rst
```diff
@@ -44,37 +47,70 @@ in `/home/system/etc/config.toml`
 Querying
 --------
 
-A current list of all available versions of all registered applications can be generated by using the ``apps`` query.
+Registered Applications
+~~~~~~~~~~~~~~~~~~~~~~~
 
-For example::
+A current list of all available versions of all registered applications can be generated by using the ``registeredApps`` query.
```
cantino: To clarify, does this include previous versions of every app, or only the latest version?

docs/ecosystem/services/app-service.rst
```diff
@@ -221,6 +297,9 @@ argument, as well as any additional arguments specified with ``args``.
 If the application immediately fails, the ``errors`` field will contain a message with the
 application's return code.
 
+If an instance of the application with the requested run level is currently running, the ``startApp``
```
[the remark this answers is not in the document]
cantino: Is there a way to stop an app? For example, if it's hanging?

services/app-service/src/monitor.rs
```diff
@@ -0,0 +1,114 @@
...
+    // Add the app information to our monitoring registry
+    insert_entry(registry.clone(), &app);
+
+    // Wait for the application to finish running
+    let status = process_handle
+        .wait()
+        .map_err(|err| AppError::MonitorError {
+            err: format!("Failed to wait for {} to finish: {:?}", app.name, err),
+        })
+        .unwrap();
```
cantino: Newbie question: does this panic the app service?

services/app-service/src/registry.rs
```diff
@@ -611,17 +637,60 @@ impl AppRegistry {
...
             }
-        } else {
-            Ok(child.id())
+            Ok(None) => {
+                let name = app_name.to_owned();
+                let pid = child.id() as i32;
+                let registry = self.monitoring.clone();
+
+                // Spawn monitor thread
+                thread::spawn(move || {
```
cantino: Newbie question: what happens in rust if a thread errors?

services/app-service/src/schema.rs
```diff
@@ -58,14 +58,26 @@ graphql_object!(QueryRoot : Context as "Query" |&self| {
...
+        Ok(result)
+    }
+
+    field running_apps(&executor,
+        name: Option<String>)
+        -> FieldResult<Vec<MonitorEntry>> as "Running Apps Query"
+    {
+        let entries = executor.context().subsystem().monitoring.lock()?;
+        let result = entries.iter().filter(|e| {
+            if name.is_some() && &e.name != name.as_ref().unwrap() {
```
cantino: Names are enforced to be unique and are unchangable?

services/app-service/src/monitor.rs
CatGarab: No, it just panics this particular thread (which exists for the sole purpose of monitoring this single app execution)

It's a specific behavior of Rust. A child thread crashing won't affect the parent

services/app-service/src/schema.rs
CatGarab: Name and version number form a unique pair, which is enforced. There is currently no way to rename a previously registered application (other than to manually uninstall and re-install/register)

CatGarab: I should add the ability to filter by version number as well...

services/app-service/src/registry.rs
cantino: When the thread panics, can we notice and log?

CatGarab: There's probably a procedural way to catch a panic, log a message, and then continue with the normal panic logic, but I haven't found it.

I should be logging my errors before all my unwraps, but I'm now realizing that I didn't :sweat:

docs/ecosystem/services/app-service.rst
CatGarab: All registered versions, active/latest or not. There's some example graphql output a few lines down from here.

The exception is if you uninstall a particular version (or an entire app). That version is removed from the registry, so won't appear in future queries

docs/ecosystem/services/app-service.rst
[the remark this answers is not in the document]
CatGarab: [not currently](https://trello.com/c/9qlhYENe/703-app-service-stopping-applications) via the app service.

you'd have to manually fetch the PID and then kill the process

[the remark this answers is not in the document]
cantino: We might want to consider adding that.

docs/ecosystem/services/app-service.rst
```diff
@@ -97,14 +133,52 @@ To list all available versions of a specific application, specify the app's name
...
             app {
                 name,
                 version
             }
         }
     }
+
+.. _running-apps:
     
+Running Applications
```
jacoffey3: Are you changing this to appStatus with this PR? Or a later one?

CatGarab: Later one

jacoffey3: Okay, then you can dismiss my review. It's n/a

PR #447, opened 2019-08-09
Adding ability to register app with custom config

CatGarab:
We previously added the ability for apps to be started via the `startApp` mutation with a custom config. This PR expands that so that a custom config can be defined as the default for a particular app, that way that config file will be used when that app's OnBoot logic is called during the app service's initialization.

### DIFF
--- a/services/app-service/src/tests/register_app.rs
+++ b/services/app-service/src/tests/register_app.rs
@@ -51,6 +51,7 @@ fn register_good() {
                 active, 
                 app {{
                     author,
+                    config,
                     name,
                     version,
                 }}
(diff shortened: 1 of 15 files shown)


CatGarab:
[OnBoot, there is currently no way to provide a custom config.](https://trello.com/c/q1ZGY73u/1037-onboot-there-is-currently-no-way-to-provide-a-custom-config)

services/app-service/src/tests/register_app.rs
```diff
@@ -151,6 +153,81 @@ fn register_good_exec() {
     assert!(path.ends_with("dummy/0.0.1/dummy2"));
 }
 
+#[test]
+fn register_good_config() {
```
cantino: What happens if the config file cannot be found?

jacoffey3: Made a card for this: https://trello.com/c/I0LEovtB/1058-config-file-failover

PR #451, opened 2019-08-15
Allow bulk telemetry over UDP

cantino:
If telemetry entries come wrapped in an array over UDP, they are all imported.

What's the best way to write a test that asserts error logs?

### DIFF
--- a/services/telemetry-service/src/udp.rs
+++ b/services/telemetry-service/src/udp.rs
@@ -16,12 +16,20 @@
 
 use kubos_telemetry_db::Database;
 use log::{error, info};
-use serde_json::{self, Value};
+use serde::Deserialize;
 use std::net::{SocketAddr, UdpSocket};
 use std::sync::{Arc, Mutex};
 
 pub struct DirectUdp {
     db: Arc<Mutex<Database>>,
+}
+
+#[derive(Debug, Deserialize)]
+struct DataPoint {
+    timestamp: Option<f64>,
+    subsystem: String,
+    parameter: String,
+    value: String,
 }
 
 impl DirectUdp {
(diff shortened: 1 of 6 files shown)


cantino:
```
error: 'cargo-fmt' is not installed for the toolchain '1.32.0-x86_64-unknown-linux-gnu'
To install, run `rustup component add rustfmt --toolchain 1.32.0-x86_64-unknown-linux-gnu`
```
?

plauche:
You'll need to rebase this off master. We recently updated the `rust_toolchain` and CI to use rust version 1.36.0. This branch is probably still trying to use 1.32.0.

services/telemetry-service/src/udp.rs
```diff
@@ -16,14 +16,22 @@
...
+use serde::Deserialize;
 use std::net::{SocketAddr, UdpSocket};
 use std::sync::{Arc, Mutex};
 
 pub struct DirectUdp {
     db: Arc<Mutex<Database>>,
 }
 
+#[derive(Debug, Deserialize)]
+struct DataPoint {
```
cantino: Should I be importing this? https://github.com/kubos/kubos/blob/e79d357c16689e37f4c89b96090a92e696f8c0b3/apis/telemetry-db-api/src/models.rs#L22

cantino:
It seems like I should also add this to our benchmark, but looking at https://github.com/kubos/kubos/blob/66a54c036b464795ebe8fa498cdd996e4991f4a0/test/benchmark/db-test/src/main.rs#L259, I don't see how it won't always be 2ms since UDP is nonblocking?

cantino:
Tested locally with 

```
echo '[{ "subsystem": "test1", "parameter": "current", "value": "2.0" }, { "subsystem": "test1", "parameter": "current", "value": "3.0" }]' | nc -q 1 -u 127.0.0.1 8001

echo '{ "subsystem": "test1", "parameter": "current", "value": "2.0" }' | nc -q 1 -u 127.0.0.1 8001
```

and a temporary `debug!("Stored {:?}", &message);`

(Which I assume I shouldn't commit?)

```
Aug 15 19:45:59 vagrant kubos-telemetry-service[9518]: Stored DataPoint { timestamp: None, subsystem: "test1", parameter: "current", value: "2.0" }
```

cantino:
Does it make any sense to add it to the benchmark, then? Since you can't really time the UDP from the client side.

PR #453, opened 2019-08-16
Service logging fix + app service OnBoot fix

CatGarab:
This PR fixes two bugs which slipped through the cracks of our CI tests:

- When the services start up, they attempt to read various config options. There is currently `ok_or()` logic to log an error message and then exit the service. `ok_or()` is "eagerly evaluated" by the Rust compiler, so the error message ends up _always_ getting logged, even if the config fetch succeeded.
  - Migrating to `ok_or_else` instead to avoid this problem
  - Problem wasn't caught by CI because we don't verify logging
- Our config parsing logic currently uses `getopts` to parse the command line options and look for the `-c {config.toml}` argument. Now that config parsing no longer silently fails and takes defaults, the init system is failing to start the app service. It tries to call `kubos-app-service -b` to trigger the OnBoot logic. The `-b` option isn't recognized by the config parser, so everything crashes
  - Updating `get_config_path` to manually look for the `-c` option, rather than by using `getopts`, so that way we can ignore any unknown arguments (under the assumption that they will be parsed by whatever entity is trying to parse a config file)
  - Updating the way the app service fetches its config file so that it a) matches all the other services and b) causes the unit test to correctly fail with the old `get_config_path` logic

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1225,7 +1225,6 @@ name = "kubos-system"
 version = "0.1.0"
 dependencies = [
  "failure 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "getopts 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.89 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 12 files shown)


cantino:
I also would like to see some logging assertions in tests.

PR #454, opened 2019-08-22
Exclusively testing Python app run failure in CI

CatGarab:
The app service currently waits for 300ms for apps to execute and return. It's a known problem that Python apps can take longer to run, depending on the host system. For CircleCI, this means that the `app_failure` test works just fine. However, when `cargo test` is run during the Vagrant provisioning process, this test almost always fails.

I'm marking it with `#[ignore]` so that it won't be tested unless explicitly requested.


CatGarab:
[app service python app test causing vagrant provisioning to fail](https://trello.com/c/wVCN1ycU/1065-app-service-python-app-test-causing-vagrant-provisioning-to-fail)

PR #455, opened 2019-08-23
Manipulating Dockerfile to shrink file Docker image

CatGarab:
Each `RUN` line in the Dockerfile is created as a separate "layer". It turns out that layers cannot make a final image smaller, they can only make it larger. As a result, it is recommended that any "remove file" commands be combined with their original "get file" command into one single `RUN` step.
Most of the changes in this PR address that paradigm.

Additionally, removing the rust toolchain docs because they're 300MB and unnecessary for CI


CatGarab:
[Shrink Docker image](https://trello.com/c/bR4gNTjj/1066-shrink-docker-image)

PR #471, opened 2019-09-10
Improving Getting Started workflow

CatGarab:
- Adding a "Next Steps" section to the local dev setup guide (all other docs already have one)
- Moving the "Running Services Locally" links to be mentioned later on in the Getting Started list. We'd like the users to go look at our tutorials first

### DIFF
--- a/docs/getting-started/index.rst
+++ b/docs/getting-started/index.rst
@@ -61,11 +61,6 @@ Using C
 If you would like to develop in C, we highly recommend you :doc:`install the SDK <../sdk-docs/sdk-installing>`
 and then refer to :doc:`../sdk-docs/sdk-c`.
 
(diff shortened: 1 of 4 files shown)


CatGarab:
[Call to Action at the bottom of all getting started docs](https://trello.com/c/oNzrBTNL/1077-call-to-action-at-the-bottom-of-all-getting-started-docs)

CatGarab:
[Getting services running locally doesn't actually tell you what services to run. We should not have people go to it first, but rather link to it from tutorials](https://trello.com/c/dnjFWsao/1079-getting-services-running-locally-doesnt-actually-tell-you-what-services-to-run-we-should-not-have-people-go-to-it-first-but-rath)

PR #472, opened 2019-09-12
Scheduler tokio + general refactor

plauche:
- Refactored `tokio` usage to give us a single runtime in a single thread. This enabled to creation of `stop` to stop an active schedule and all scheduled tasks. Addresses naive implementation found in #465.
- Implementing starting or restarting the active schedule upon activating a schedule.
- Extracted `config.rs` for `ScheduleConfig` related structures and parsing function.
- Extracted `file.rs` for `ScheduleFile` related things, and functions which deal with manipulating on disk schedule files.
- Distilled `scheduler.rs` down to functions and structures solely related to running the actual schedule and executing app tasks.

### DIFF
--- a/services/scheduler-service/src/config.rs
+++ b/services/scheduler-service/src/config.rs
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2019 Kubos Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//!
+//! Definitions and parsing instructions for Schedule Configurations
+//!
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::time::Duration;
+
+// Configuration used for execution of an app
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ScheduleApp {
+    pub name: String,
+    pub args: Option<Vec<String>>,
+    pub config: Option<String>,
+    pub run_level: Option<String>,
+}
+
+// Configuration used to schedule app execution
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ScheduleTask {
+    // Start delay specified in HH:mm:ss format
+    // Used by init and recurring tasks
+    pub delay: Option<String>,
+    // Details of the app to be executed
+    pub app: ScheduleApp,
+}
+
+impl ScheduleTask {
+    // Parse delay duration from delay field
+    pub fn get_duration(&self) -> Result<Duration, String> {
+        if let Some(delay) = &self.delay {
+            let mut task_delay: Vec<&str> = delay.split(':').collect();
+            let delay_s: u64 = task_delay.pop().unwrap_or("").parse().unwrap_or(0);
+            let delay_m: u64 = task_delay.pop().unwrap_or("").parse().unwrap_or(0) * 60;
+            let delay_h: u64 = task_delay.pop().unwrap_or("").parse().unwrap_or(0) * 3600;
+            Ok(Duration::from_secs(delay_s + delay_m + delay_h))
+        } else {
+            Err("No delay defined for task".to_owned())
+        }
+    }
+}
+
+// Schedule configuration information - the guts of the actual schedule
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ScheduleConfig {
+    pub init: Option<HashMap<String, ScheduleTask>>,
+    pub one_time: Option<HashMap<String, ScheduleTask>>,
+}
--- a/services/scheduler-service/src/file.rs
+++ b/services/scheduler-service/src/file.rs
@@ -1,0 +1,176 @@
+/*
+ * Copyright (C) 2019 Kubos Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+//!
+//! Definitions and functions concerning the manipulation of schedule files
+//!
+
+use chrono::{DateTime, Utc};
+use log::{info, warn};
+use std::fs;
+use std::os::unix::fs::symlink;
+use std::path::{Path, PathBuf};
+
+// Descriptive information about a Schedule File
+#[derive(Debug, GraphQLObject)]
+pub struct ScheduleFile {
+    pub contents: String,
+    pub path: String,
+    pub name: String,
+    pub time_registered: String,
+    pub active: bool,
+}
+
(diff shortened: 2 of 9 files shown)


services/scheduler-service/src/config.rs
```diff
@@ -0,0 +1,64 @@
...
+    pub config: Option<String>,
+    pub run_level: Option<String>,
+}
+
+// Configuration used to schedule app execution
+#[derive(Debug, Serialize, Deserialize)]
+pub struct ScheduleTask {
+    // Start delay specified in HH:mm:ss format
+    // Used by init and recurring tasks
+    pub delay: Option<String>,
```
cantino: Do you think we should store this as a `Duration` or `uint` seconds instead and parse / validate it before making the struct?

services/scheduler-service/src/file.rs
```diff
@@ -0,0 +1,176 @@
...
+pub fn import_schedule(scheduler_dir: &str, path: &str, name: &str) -> Result<(), String> {
+    info!("Importing new schedule '{}': {}", name, path);
+    let schedule_dest = format!("{}/{}.json", scheduler_dir, name);
+    fs::copy(path, schedule_dest).map_err(|e| format!("Schedule copy failed: {}", e))?;
+    Ok(())
+}
+
+// Make an existing schedule file the active schedule file
+// TODO: Should this kill the schedule and rerun scheduler.start()?
+// TODO: What if this is the active schedule?? What would happen?
```
cantino: Probably should return early if it's already active?

services/scheduler-service/src/file.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    fs::rename(&new_active_path, &active_path)
+        .map_err(|e| format!("Failed to copy over new active symlink: {}", e))?;
+
+    info!("Activated schedule {}", name);
+    Ok(())
+}
+
+// Remove an existing schedule file from the schedules directory
+// TODO: What happens if this is the active schedule?
```
cantino: Should error on active schedule.

services/scheduler-service/src/file.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    fs::remove_file(&sched_path)
+        .map_err(|e| format!("Failed to remove schedule {}.json: {}", name, e))?;
+
+    info!("Removed schedule {}", name);
+    Ok(())
+}
+
+// Retrieve information on the active schedule file
+pub fn get_active_schedule(scheduler_dir: &str) -> Option<ScheduleFile> {
```
cantino: Should this return a `Result`?

services/scheduler-service/src/scheduler.rs
```diff
@@ -14,18 +14,131 @@
...
+    let name = task.app.name.clone();
+    let config = task.app.config.clone();
+    let args = task.app.args.clone();
+    let runlevel = task.app.run_level.clone().unwrap_or("onBoot".to_owned());
+
+    let duration = match task.get_duration() {
+        Ok(d) => d,
+        Err(e) => {
+            error!("Failed to parse task delay duration: {}", e);
+            panic!("Failed to parse task delay duration: {}", e)
```
cantino: Is this crashing the main thread now?

services/scheduler-service/src/scheduler.rs
```diff
@@ -14,18 +14,131 @@
...
+    };
+
+    let when = Instant::now() + duration;
+
+    Delay::new(when)
+        .and_then(move |_| {
+            info!("Start app {}", name);
+            let mut query_args = format!("runLevel: \"{}\", name: \"{}\"", runlevel, name);
+            if let Some(config) = config {
+                query_args.push_str(&format!(", config: \"{}\"", config));
```
cantino: *Reminder that we need to switch to GraphQL variables and get away from string injection.*

services/scheduler-service/src/scheduler.rs
```diff
@@ -40,139 +153,49 @@ impl Scheduler {
...
-        if let Some(init) = active_config.init {
-            for (name, task) in init {
-                let task_app_url = apps_service_url.clone();
-                info!("Scheduling {}", name);
-                if let Err(e) = task.schedule(task_app_url) {
-                    error!("Failed to schedule task {}: {:?}", name, e);
+        let handle = thread::spawn(move || {
+            tokio::run(lazy(move || {
+                if let Some(init) = active_config.init {
+                    dbg!(&init);
```
cantino: `dbg!`

services/scheduler-service/src/scheduler.rs
```diff
@@ -40,139 +153,49 @@ impl Scheduler {
...
-    pub fn remove_schedule(&self, name: &str) -> Result<(), String> {
-        info!("Removing schedule {}", name);
-        let sched_path = format!("{}/{}.json", self.scheduler_dir, name);
-
-        fs::remove_file(&sched_path)
-            .map_err(|e| format!("Failed to remove schedule {}.json: {}", name, e))?;
+    // Kill the scheduler thread and hard stop the active tokio runtime
+    pub fn stop(&self) -> Result<(), String> {
+        let mut my_handle = self.scheduler_thread.lock().unwrap();
+        *my_handle = None;
```
cantino: *newbie question*: is this triggering Drop? Do we have a test that shows that future schedules do indeed now not get run?

services/scheduler-service/src/config.rs
plauche: That would be nice, but it would mean either implementing a custom parser for `serde` or custom parsing instead of `serde`.  I'm ok with leveraging `serde` for parsing magic as is, and pulling out the duration later.

services/scheduler-service/src/scheduler.rs
plauche: It wasn't crashing the main thread, but it was disrupting the parsing of any tasks afterwards :(.

services/scheduler-service/src/scheduler.rs
plauche: It was dropping the thread, but not stopping the runtime. Reworked to properly stop the runtime and added a test to verify.

services/scheduler-service/src/scheduler.rs
```diff
@@ -40,139 +165,61 @@ impl Scheduler {
...
-            }
-        }
-
-        Ok(())
-    }
+                Ok(())
+            }));
+            // Wait on the stop message before ending the runtime
+            rx.recv().unwrap();
+            runner.shutdown_now().wait().unwrap();
```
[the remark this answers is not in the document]
plauche: The call to `shutdown_now` just returns a future, it does not block until the runtime is done. Calling `wait()` here blocks until that future is complete and the runtime has fully shutdown.

services/scheduler-service/src/scheduler.rs
```diff
@@ -40,139 +165,61 @@ impl Scheduler {
...
 
         let apps_service_url = apps_service_config
             .hosturl()
             .ok_or_else(|| "Failed to fetch app service url".to_owned())?;
 
-        let active_schedule = self
-            .get_active_schedule()
+        let active_schedule = get_active_schedule(&self.scheduler_dir)
             .ok_or_else(|| "Failed to fetch active schedule".to_owned())?;
         let active_config: ScheduleConfig = serde_json::from_str(&active_schedule.contents)
```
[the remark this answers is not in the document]
plauche: Yea that is a good idea. I went ahead and made the change in #477 because stuff was already moving around and it would cause less conflicts that way. 

https://github.com/kubos/kubos/pull/477/files#diff-b506f4bc6642789ffb529b5197a44be8R30

services/scheduler-service/src/scheduler.rs
```diff
@@ -40,139 +165,61 @@ impl Scheduler {
...
 
-        if let Some(init) = active_config.init {
-            for (name, task) in init {
-                let task_app_url = apps_service_url.clone();
-                info!("Scheduling {}", name);
-                if let Err(e) = task.schedule(task_app_url) {
-                    error!("Failed to schedule task {}: {:?}", name, e);
+        let (tx, rx) = channel::<()>();
+        let handle = thread::spawn(move || {
+            let mut runner = Runtime::new().unwrap();
```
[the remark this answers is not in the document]
plauche: Fixed in #477 

https://github.com/kubos/kubos/pull/477/files#diff-9e6fe68762c86a7be575361019715140R189

PR #476, opened 2019-09-18
Fixing ISIS antenna structure ownership

CatGarab:
Primary change: The ISIS antenna service now initiates the `AntS` structure inside a mutex. Previously we were allowing it to be cloned, which lead to the HTTP server cloning the structure, doing its GraphQL query work, then dropping the clone. This clone triggered the `AntS::Drop()` logic, which closes the I2C connection, preventing any further communication with the underlying device.

Auxiliary changes: Bringing code up to rust 1.37 standards, disallowing warnings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1009,10 +1009,9 @@ name = "isis-ants"
 version = "0.1.0"
 dependencies = [
  "isis-ants-api 0.1.0",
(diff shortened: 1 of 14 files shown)


CatGarab:
[Fix ISIS Antenna Service](https://trello.com/c/ijyr3C8m/1115-fix-isis-antenna-service)

PR #478, opened 2019-09-26
Updating default config file location and behavior

CatGarab:
The default config.toml file is being moved from `/home/system/etc/config.toml` to `/etc/kubos-config.toml`. This allows the config file to be recovered via the usual OS recovery process.

As a result, users who wish to create custom config options must use a custom config file living in the user data partition and must specify the `-c {path}/{to}/{config}` option when running services and applications.

The PR is created in conjunction with kubos/kubos-linux-build#191


CatGarab:
[Config File Failover](https://trello.com/c/I0LEovtB/1058-config-file-failover)

jacoffey3:
SO MANY CONFIGS. But actually, what would be a helpful review here. I generally trust your ctrl+r skills

CatGarab:
@jacoffey3 There are a few doc section updates/additions which should be reviewed

jacoffey3:
> Ok now that the `default_config.toml` has been updated, it gives an easy map of the core services and their respective ports, and shows some gaps. What is supposed to live in `801x`, `802x`, and `804x`?
> 
> It would also be good if we added these default port values to our service docs, or to have one doc where you can see all of the port mappings.

`801x` will be the scheduler. `802x` and `804x`are not assigned to anything in particular.

plauche:
Can we move all of the port mappings down, so that the available ranges are grouped at the high end of the numbers? I think that would make them `806x` and `807x`? It feels a bit haphazard to have our unassigned ranges appear in arbitrary places.

jacoffey3:
> Can we move all of the port mappings down, so that the available ranges are grouped at the high end of the numbers? I think that would make them `806x` and `807x`? It feels a bit haphazard to have our unassigned ranges appear in arbitrary places.


I guess? I don't have anything against it, but they are not available to anyone but internally. They are reserved for Core Services only

CatGarab:
[Lock port locations for Core Services](https://trello.com/c/yuY6c4kN/1097-lock-port-locations-for-core-services)

PR #480, opened 2019-09-30
Changing uninstall logic to nicely kill running app first

CatGarab:
Currently, if you uninstall a running application, we immediately try to kill it with SIGKILL (the very harsh, no cleanup signal). Updating this logic to first send SIGTERM (graceful kill), followed by SIGKILL two seconds later.


CatGarab:
[Send SIGTERM when uninstalling instead of SIGKILL at first, sleep for 2 seconds, then issue SIGKILL if it hasn't stopped.](https://trello.com/c/f59ZH7kK/1082-send-sigterm-when-uninstalling-instead-of-sigkill-at-first-sleep-for-2-seconds-then-issue-sigkill-if-it-hasnt-stopped)

jacoffey3:
Docs look good 👍

cantino:
Makes sense to me. Do we do the same 2s logic when you stop an app?

CatGarab:
@cantino When you use the `killApp`, it sends whatever signal you specify. By default, it's the SIGTERM signal. Because it's already a manual process, we can expect that users will send the SIGKILL signal if they really need their app to die.

services/app-service/src/registry.rs
```diff
@@ -757,3 +755,17 @@ impl AppRegistry {
         })
     }
 }
+
+fn uninstall_kill(pid: i32) -> Result<(), nix::Error> {
+    let pid = Pid::from_raw(pid);
+    signal::kill(pid, Some(signal::Signal::SIGTERM))?;
```
plauche: I wonder if we should still attempt to `SIGKILL` even if this one fails.

CatGarab: I don't think it would change anything. Looks like there are three reasons why `kill()` could fail:
- Invalid signal
- Process doesn't have permissions to be sending signals
- The PID doesn't exist

PR #481, opened 2019-10-04
Commonizing service behavior

CatGarab:
- `tools/default_config.toml` -> `tools/local_config.toml`
- Updating `local_config.toml` to have a relative app service registry dir, so that it can easily be run locally (the telemetry and file service configs already do this)
- Adding error messages to all remaining config load/parse errors


CatGarab:
[Make all Core Services behave the same way in respect to storage locations and behave the same with regards to config files](https://trello.com/c/BG2XPBzZ/1112-make-all-core-services-behave-the-same-way-in-respect-to-storage-locations-and-behave-the-same-with-regards-to-config-files)

PR #484, opened 2019-10-10
File service + client fixes

plauche:
- File service now logs a message if config file is missing
- File service and file client will now run on compatible, default ip/port values if not specified
- Adjusted default chunk size in file service to match client

### DIFF
--- a/clients/kubos-file-client/src/main.rs
+++ b/clients/kubos-file-client/src/main.rs
@@ -197,7 +197,8 @@ fn main() {
             Arg::with_name("host_port")
                 .short("-P")
                 .help("UDP port that the file transfer service will send responses to")
-                .takes_value(true),
+                .takes_value(true)
+                .default_value("8080"),
         )
         .arg(
             Arg::with_name("remote_ip")
--- a/services/file-service/src/lib.rs
+++ b/services/file-service/src/lib.rs
@@ -47,8 +47,8 @@ pub fn recv_loop(config: &ServiceConfig) -> Result<(), failure::Error> {
 
     // Get the chunk size to be used for transfers
     let chunk_size = match config.get("chunk_size") {
-        Some(val) => val.as_integer().unwrap_or(4096),
-        None => 4096,
+        Some(val) => val.as_integer().unwrap_or(1024),
+        None => 1024,
     } as usize;
 
     let hold_count = match config.get("hold_count") {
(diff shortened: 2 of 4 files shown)


plauche:
[Update File Transfer Service to not fail silently if there isn't a configured downlink IP and port](https://trello.com/c/efmldKjT/1137-update-file-transfer-service-to-not-fail-silently-if-there-isnt-a-configured-downlink-ip-and-port)

apis/system-api/src/config.rs
```diff
@@ -91,7 +91,11 @@ impl Config {
     /// `name` - Category name used as a key in the config file
     /// `path` - Path to configuration file
     pub fn new_from_path(name: &str, path: String) -> Result<Self, Error> {
-        parse_config_file(name, path)
+        if std::path::Path::new(&path).is_file() {
```
CatGarab: Why was this new logic necessary?

clients/kubos-file-client/src/main.rs
```diff
@@ -197,7 +197,8 @@ fn main() {
             Arg::with_name("host_port")
                 .short("-P")
                 .help("UDP port that the file transfer service will send responses to")
-                .takes_value(true),
+                .takes_value(true)
+                .default_value("8080"),
```
CatGarab: The readme will need to be updated to show that this is now an optional argument.

It might also be mentioned somewhere in the docs?

services/file-service/src/lib.rs
```diff
@@ -60,15 +60,15 @@ pub fn recv_loop(config: &ServiceConfig) -> Result<(), failure::Error> {
     let downlink_port = config
         .get("downlink_port")
         .and_then(|i| i.as_integer())
-        .expect("Downlink port not found") as u16;
+        .unwrap_or(8080) as u16;
```
CatGarab: You're creating two different behaviors between the IP and port params.

For the port, if the `downlink_port` param doesn't exist, or can't be parsed, we take the default value.
For the IP address, if the param doesn't exist we take the default, but if it can't be parsed we return an error.

apis/system-api/src/config.rs
plauche: Hmm I'm not sure why I though that was needed...but its not.

clients/kubos-file-client/src/main.rs
jacoffey3: It was already documented as optional, so this makes it match the docs

services/file-service/src/lib.rs
```diff
@@ -60,15 +60,15 @@ pub fn recv_loop(config: &ServiceConfig) -> Result<(), failure::Error> {
...
-        .expect("Downlink IP not found")
-        .as_str()
-        .and_then(|str| Some(str.to_owned()))
-        .expect("Downlink IP not found");
+    let downlink_ip_op = config.get("downlink_ip");
+    let downlink_ip = match downlink_ip_op {
+        Some(ip) => ip.as_str().and_then(|ip| Some(ip.to_owned())),
+        None => Some("127.0.0.1".to_owned()),
+    }
+    .ok_or_else(|| failure::format_err!("Failed to parse downlink IP address"))?;
```
jacoffey3: Will it still panic if it isn't able to assign the IP? Or will it show this message?

clients/kubos-file-client/src/main.rs
CatGarab: Where? The only place I can actually find where it's mentioned is the readme, which says that it's a required argument

CatGarab: It's also mentioned (as required) in the [file transfer tutorial](https://docs.kubos.com/1.18.0/tutorials/file-transfer.html#syntax)

services/file-service/src/lib.rs
plauche: It will now no longer panic if the IP cannot be found or parsed from the config file. However, if the assigned IP is already in use, then the service will issue an error message and panic.

PR #489, opened 2019-10-18
Fixing relative app-registry paths

plauche:
There is an issue creating symlinks in the `active` dir when the base `apps_dir` is a relative path. This ensures that `apps_dir` is always an absolute path.


plauche:
[Fixing relative app registry paths](https://trello.com/c/i8gNKFKr/1150-fixing-relative-app-registry-paths)

PR #490, opened 2019-10-23
Expanding shell client to allow remote execution of single commands

plauche:


plauche:
[App to execute arbitrary shell commands](https://trello.com/c/pmFX8PBo/1154-app-to-execute-arbitrary-shell-commands)

clients/kubos-shell-client/src/main.rs
```diff
@@ -36,6 +36,46 @@ fn start_session(channel_proto: &ChannelProtocol) -> Result<(), Error> {
     Ok(())
 }
 
+fn run_command(channel_proto: &ChannelProtocol, command: &str) -> Result<(), Error> {
+    let channel_id = channel_protocol::generate_channel();
+
+    let mut split_command: Vec<String> = command.split(' ').map(|s| s.to_owned()).collect();
```
cantino: Is this going to split whitespace inside of strings?

Also, things like "ENV=foo command arg1 arg2" are going to give the wrong `base_command`.

clients/kubos-shell-client/src/main.rs
```diff
@@ -36,6 +36,46 @@ fn start_session(channel_proto: &ChannelProtocol) -> Result<(), Error> {
...
+
+    println!("Starting shell session -> {}", channel_id);
+
+    channel_proto.send(&shell_protocol::messages::spawn::to_cbor(
+        channel_id,
+        &base_command,
+        Some(&split_command),
+    )?)?;
+
+    loop {
```
cantino: We have no idea when the command has exited?

clients/kubos-shell-client/README.md
```diff
@@ -6,12 +6,14 @@ This client program can be used to test communication with the Kubos shell servi
 
 To build and run the client program, run the following command from this folder:
 
-    cargo run -- (start|list|join|kill) [config-options]
+    cargo run -- (start|run|list|join|kill) [config-options]
```
jacoffey3:
```suggestion
    cargo run --bin kubos-shell-client -- (start|run|list|join|kill) [config-options]
```

clients/kubos-shell-client/README.md
```diff
@@ -6,12 +6,14 @@ This client program can be used to test communication with the Kubos shell servi
...
     
 Required arguments:
 
 - Operation to perform
     - `start` - Start a new shell session
+    - `run -c "{command}"` - Runs a one-shot remote command
+        - `command` - Remote command and any required arguments
```
jacoffey3:
```suggestion
        - `command` - Remote command and any required arguments. *Must* be enclosed in `"`s.  If the command is invalid, the operation will timeout. You will not get an error message. 
```

clients/kubos-shell-client/src/main.rs
plauche: We do! I look for ``Message::Exit`` in the match, which indicates that the command has exited.

cantino: Ah, got it!

clients/kubos-shell-client/src/main.rs
plauche: It does split the white space inside of strings, but it all ends up ok in the end haha. The args vec is passed along and consumed by a call to [`Command::args`](https://doc.rust-lang.org/std/process/struct.Command.html#method.args). The list appears to end up effectively combined into one arg or string and given for consumption to the spawned process. I did some testing and did not run into behaviour issues. 

Two caveats:
- Passing in ENV variable won't work because we are directly executing the `command`, not running it though a shell session.
- Also because we are not running in an actual shell session, we can't do anything fancy like pipe chained commands `ls /home/ryan | grep Development`

@jacoffey3 do we want to support richer shell commands, or are we wanting a simpler execution of straight binaries?

plauche: It was simple enough to implement execution via `/bin/sh` instead of direct execution. I think this is the better way, as it gives us better errors when commands are invalid/incorrect, allows more complex commands, and makes the passing of command + args more straightforward.

PR #492, opened 2019-10-28
Pinning pkg-config crate version

plauche:
Pinning the version of `pkg-config` to `0.3.15` to avoid introduction of `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` default usage by `0.3.16`.

The `PKG_CONFIG_ALLOW_SYSTEM_CFLAGS` flag is currently breaking the kubos linux build.


plauche:
@cantino good call, added a note about why it's there.

PR #500, opened 2019-11-07
App service accepts archives

cantino:
Allow a `.tgz` file to be provided instead of a directory to the app service's register mutation. The binary and manifest must be in the root of the tarball (not in a nested directory), and detection is in place to warn if the user makes a mistake.

I've tested in the SDK and on my Mac. I'd appreciate if someone could test on hardware.


services/app-service/src/registry.rs
```diff
@@ -219,18 +224,81 @@ impl AppRegistry {
             return Err(AppError::RegisterError {
                 err: format!("{} does not exist", path),
             });
-        }
+        } else if !app_path.is_dir() {
+            match app_path.extension().and_then(OsStr::to_str) {
```
cantino: If I were to move some of this into a separate method or file, where would you suggest it should go?

jacoffey3:
Docs!

cantino:
Docs added.

docs/tutorials/app-register.rst
```diff
@@ -108,7 +108,8 @@ It has the following schema::
         }
      }
      
-The ``path`` input parameter specifies the directory where the application and manifest files reside.
+The ``path`` input parameter specifies the path to either a directory where the application and manifest files reside,
+or a the path to a ``.tgz`` archive file containing the application and manifest in its root.
```
plauche: nit: `or the path`

services/app-service/src/registry.rs
plauche: Yea this function is becoming a monster. You could split out a few private functions in this file. I potentially see an `extract_archive` from 230:279 and `validate_application` from 223:328.

docs/tutorials/app-register.rst
cantino: That's not a nit, that's egregious! Thanks. :)

services/app-service/src/registry.rs
cantino: I did `extract_archive` but not `validate_application` for now. I have `extract_archive` doing implicit returns, which I think is readable enough.

PR #516, opened 2020-03-05
[WIP] Very basic beacon app in Rust

cantino:


apps/rust-beacon-app/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+    beacon.la5 = la5.and_then(|v| v.as_f64());
+    beacon.la15 = la15.and_then(|v| v.as_f64());
+    beacon.root_avail = available_space_for_mount(mounts, "/");
+    beacon.home_avail = available_space_for_mount(mounts, "/home");
+    beacon.envar_avail = available_space_for_mount(mounts, "/envar");
+
+    Ok(beacon)
+}
+
+fn available_space_for_mount(mounts_value: Option<&Vec<Value>>, path: &str) -> Option<f64> {
```
cantino: @plauche I'd love suggestions on how to re-write this method to be cleaner with the serde_json Value objects. It's trying to extract the matching entry from rows like

```
[
  {
    "fsMountedOn": "/",
    "avail": 12345,
    "something": "else"
  },
  {
    "fsMountedOn": "/home",
    "avail": 1234567,
    "something": "else"
  },
  ...
]
```

jacoffey3:
Regarding the issues mentioned - 

- It actually is not only okay, but encouraged that we predict the layout of the filesystem. Since we are controlling it, we can take advantage of that being known. 
- Similarly to the above, the port/ip can be considered "known" as well, although it should likely be reading it from the system's `config.toml`. 
- Very much agree on the packet format. We can control maximum size easily right now, but you're right it'll get unwieldy fast.

apps/rust-beacon-app/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+    // This option will be processed by the system-api crate when a service query is run
+    opts.optflagopt(
+        "c",
+        "config",
+        "System config file which should be used",
+        "CONFIG",
+    );
+    opts.optflag("h", "help", "Print this help menu");
+
+    let beacon = get_beacon_information()?;
```
plauche: I wonder if the `get_beacon_information` should always return a `Beacon` and just encode any `Error` inside of the `Beacon` so that the error state is transmitted?

cantino:
Should we be reporting info about any other mounts beyond /, /home_avail, and /envar?

apps/rust-beacon-app/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+    let beacon_port = host_parts.next().unwrap_or_else(|| {
+        error!("Failed to lookup beacon host port. Using default 8500");
+        "8500".to_owned()
+    });
+    let downlink = format!("{}:{}", beacon_ip, beacon_port);
+
+    // Bind a socket for the host IP
+    let local_socket = UdpSocket::bind("0.0.0.0:0")?;
+
+    // Send our distress message
```
plauche: :sos:

apps/rust-beacon-app/src/main.rs
plauche: Hmm yes I'll have to think about that.

PR #528, opened 2020-03-27
Added new options for controlling the file protocol

plauche:
New control options:
- `hash_chunk_size` - Controls the chunk size used when hashing the file.
- `inter_chunk_delay` - Controls the delay between transmission of individual chunks
- `max_chunks_transmit` - Controls maximum number of chunks to transmit before waiting on response

Also:
- Cleaned up kubos-file-client a bit.
- Renamed original `chunk_size` to `transfer_chunk_size`

### DIFF
--- a/clients/kubos-file-client/src/main.rs
+++ b/clients/kubos-file-client/src/main.rs
@@ -24,116 +24,89 @@ use std::time::Duration;
 
 #[allow(clippy::too_many_arguments)]
 fn upload(
-    host_ip: &str,
-    host_port: u16,
-    remote_addr: &str,
+    protocol_instance: FileProtocol,
     source_path: &str,
     target_path: &str,
-    prefix: Option<String>,
-    chunk_size: usize,
-    hold_count: u16,
 ) -> Result<(), failure::Error> {
-    let f_config = FileProtocolConfig::new(prefix, chunk_size, hold_count);
-    let f_protocol =
-        FileProtocol::new(&format!("{}:{}", host_ip, host_port), remote_addr, f_config);
-
     info!(
         "Uploading local:{} to remote:{}",
         &source_path, &target_path
     );
 
     // Copy file to upload to temp storage. Calculate the hash and chunk info
-    let (hash, num_chunks, mode) = f_protocol.initialize_file(&source_path)?;
+    let (hash, num_chunks, mode) = protocol_instance.initialize_file(&source_path)?;
 
     // Generate channel id for transaction
-    let channel = f_protocol.generate_channel()?;
+    let channel = protocol_instance.generate_channel()?;
 
     // Tell our destination the hash and number of chunks to expect
-    f_protocol.send_metadata(channel, &hash, num_chunks)?;
+    protocol_instance.send_metadata(channel, &hash, num_chunks)?;
 
     // Send export command for file
-    f_protocol.send_export(channel, &hash, &target_path, mode)?;
+    protocol_instance.send_export(channel, &hash, &target_path, mode)?;
 
     // Start the engine to send the file data chunks
-    f_protocol.message_engine(
-        |d| f_protocol.recv(Some(d)),
+    protocol_instance.message_engine(
+        |d| protocol_instance.recv(Some(d)),
         Duration::from_secs(2),
         &State::Transmitting,
     )?;
     Ok(())
 }
 
-#[allow(clippy::too_many_arguments)]
 fn download(
-    host_ip: &str,
-    host_port: u16,
-    remote_addr: &str,
+    protocol_instance: FileProtocol,
     source_path: &str,
     target_path: &str,
-    prefix: Option<String>,
-    chunk_size: usize,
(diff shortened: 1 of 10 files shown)


clients/kubos-file-client/src/main.rs
```diff
@@ -188,50 +161,82 @@ fn main() {
         )
         .arg(
             Arg::with_name("host_ip")
-                .short("-h")
                 .help("IP address of the local host to use")
+                .long("host_ip")
+                .short("-h")
```
CatGarab: What do you think about using `-l` for the host IP (short for "local")? That way `-h` can be used for the help text

libs/file-protocol/src/protocol.rs
```diff
@@ -409,13 +428,21 @@ impl Protocol {
...
 
-    // Send all requested chunks of a file to the remote destination
+    /// Send all requested chunks of a file to the remote destination
+    ///
+    /// # Arguments
+    /// * channel_id - ID of channel to communicate over
+    /// * hash - Hash of file corresponding to chunks
+    /// * chunks - List of chunk ranges to transmit
+    /// * inter_chunk_delay - Delay duration between each chunk transmission
+    /// * max_chunks_transmit - Maximum number of chunks to actually transmit
```
CatGarab: `inter_chunk_delay` and `max_chunks_transmit` aren't actually function args, they're pulled out of the config

clients/kubos-file-client/src/main.rs
plauche: Yes I like that! The naming of those args has gotten a bit gnarly haha.

plauche: Well now that I look at the ip/port args again I have another idea. What if we did away with the short args and made them all long?

Right now we have:
- `-l` for `host_ip`
- `-P` for `host_port`
- `-r` for `remote_ip`
- `-p` for `remote_port`

What if we just had `--host-ip` `--host-port` `--remote-ip` and `--remort-port`. They are a bit longer but they are less confusing.

docs/ecosystem/services/file.rst
```diff
@@ -164,13 +164,20 @@ defined in the system's :doc:`config.toml <../services/service-config>` file:
         - ``timeout`` - `Default: 2.` The length of time, in seconds, for which the service
           should wait for new messages from the client once a file protocol transaction has
           been started
-        - ``chunk_size`` - `Default: 4096.` Each file is broken up into equally sized
+        - ``transfer_chunk_size`` - `Default: 1024.` Each file is broken up into equally sized
           chunks prior to transfer. This option specifies the size of those chunks
           in bytes.
+        - ``hash_chunk_size`` - `Default: 2048.` Each file is broken up into equally sized
```
jacoffey3, line 170: Why is it chunked separately for the hash and for the transfer?

docs/ecosystem/services/file.rst
```diff
@@ -164,13 +164,20 @@ defined in the system's :doc:`config.toml <../services/service-config>` file:
         - ``timeout`` - `Default: 2.` The length of time, in seconds, for which the service
```
jacoffey3, line 164: While you're here, please remove the line in the section above saying "it'll be configurable in a future release", since it is already

docs/ecosystem/services/file.rst
```diff
@@ -164,13 +164,20 @@ defined in the system's :doc:`config.toml <../services/service-config>` file:
...
           in bytes.
+        - ``hash_chunk_size`` - `Default: 2048.` Each file is broken up into equally sized
+          chunks which are used to calculate the file's hash. This option specifies the size
+          of those chunks.
         - ``hold_count`` - `Default: 5.` The number of times the protocol waits for
           a new message before ending the transaction.
         - ``downlink_ip`` - `Required` The IP address that the file service responds to.
         - ``downlink_port`` - `Required` The port that the file service responds to.
+        - ``inter_chunk_delay`` - `Default: 1` The delay, in milliseconds, taken 
+            between the transmission of each chunk.
```
jacoffey3, line 178:
```suggestion
            between the transmission of each chunk. This is to allows manual flow 
            control to tailor performance to your link behavior.
```

clients/kubos-file-client/src/main.rs
CatGarab: I'm lazy, so I'd definitely prefer to be able to type fewer characters :P

docs/ecosystem/services/file.rst
plauche: I found in my testing that the hashing is the biggest bottleneck of the import process. By increasing the chunk size used when hashing, the import time can be greatly decreased. This control allows us to increase the hash size/decrease import time as allowed by system resources.

plauche: Though I just remembered that I need to make the change on the export side to use the hash chunking size.

jacoffey3: Ahh that makes a lot of sense!

PR #537, opened 2020-03-31
Updating app and telemetry service unit tests to use Drop

CatGarab:
The app and telemetry service unit tests currently use a custom `teardown` function to clean up at the end of a test. Converting them to use the standard Rust `drop` function that way they don't have to be explicitly called


CatGarab:
[Unit tests: change `fixture.teardown()` to a Rust `Drop` definition](https://trello.com/c/jA8p4h3g/1072-unit-tests-change-fixtureteardown-to-a-rust-drop-definition)

PR #540, opened 2020-04-02
Increasing scheduler 'start app' response wait time to match app service

CatGarab:
When kicking off an app, the app service [waits 300ms](https://github.com/kubos/kubos/blob/master/services/app-service/src/registry.rs#L722) to see if the app immediately returns before returning its GraphQL response.

The scheduler is currently only waiting 200ms to hear back from the app service.

The PR bumps the scheduler's wait time to better match the app service's behavior


CatGarab:
[Scheduler times out when starting app](https://trello.com/c/w9zPlKgC/1261-scheduler-times-out-when-starting-app)

PR #548, opened 2020-05-21
Add a second type paramter allowing seperate Read/Wrtie Connections in the CommsControlBlock

grantperry:
This the CommsControlBlock can still be used in the same way as before
by setting the type parameter of WriteConnection and ReadConnection to
the same type and using one Arc<Mutex<..>>. The allows the read/write
threads to mutually lock each-other like before.

In the other case, Two separate types can be used, two separate
Arc<Mutex<..>> (Possibly don't need the mutex) allow the read/write
threads to operate independently of each-other.

Technically this is a breaking change so maybe bump the minor version of libs/comms-service?

### DIFF
--- a/libs/comms-service/src/service.rs
+++ b/libs/comms-service/src/service.rs
@@ -36,15 +36,15 @@ pub type WriteFn<Connection> =
 
 /// Struct that holds configuration data to allow users to set up a Communication Service.
 #[derive(Clone)]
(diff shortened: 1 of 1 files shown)


plauche:
@grantperry I'm going to merge this into the uclibc branch for now. If we decide to bring it into master later then we'll do a version number bump as well.

PR #551, opened 2020-10-10
Add some clarification to the colors used in the "Kubos Design" diagram

MoralCode:
document the response to a question answered by @tbrowder1 on slack regarding the meaning of colors in the diagram and also adds alt text to the figure.


plauche (approved):
Looks good! Thanks for this.

Issue #550, opened 2020-06-28
tests::registry_start_app::start_app_* fails because of exec format error


minecraft2048:
**Description**
tests::registry_start_app::start_app_* fails because of exec format error, which is caused by leading spaces in the test script tiny-app 

**System Details**
- Host PC running Ubuntu 20.04 LTS
- Kubos git master

**Severity**
- Mild

**To Reproduce**
1. Run `cargo test --package kubos-app-service` in kubos directory
2. See test fails

**Bug Output**
```
byakuya@seireitei ~/D/kubos (master) [101]> cargo test --package kubos-app-service
   Compiling kubos-app-service v0.1.0 (/home/byakuya/Development/kubos/services/app-service)
    Finished dev [unoptimized + debuginfo] target(s) in 5.64s
     Running target/debug/deps/kubos_app_service-702543978e187954

running 29 tests
test tests::ping ... ok
test tests::register_app::register_bad_name ... ok
test tests::register_app::register_archive_no_extension ... ok
test tests::register_app::register_good ... ok
test tests::register_app::register_bad_path ... ok
test tests::register_app::register_duplicate ... ok
test tests::register_app::register_archive_wrong_extension ... ok
test tests::register_app::register_no_author ... ok
test tests::register_app::register_good_exec ... ok
test tests::registry_start_app::start_app_bad ... ok
test tests::register_app::register_no_active ... ok
test tests::register_app::register_good_config ... ok
test tests::registry_start_app::start_app_fail ... ok
test tests::register_app::register_no_version ... ok
test tests::register_app::register_no_manifest ... ok
test tests::register_app::register_no_name ... ok
test tests::registry_test::serialize_entry ... ok
test tests::registry_test::empty_apps_dir_empty_reg ... ok
test tests::registry_test::invalid_apps_dir_empty_reg ... ok
test tests::register_app::register_archive_bad_structure ... ok
test tests::register_app::register_good_archive ... ok
test tests::registry_test::custom_apps_dir ... ok
test tests::registry_start_app::start_app_good ... FAILED
test tests::set_version::set_version_bad ... ok
test tests::upgrade_app::upgrade_bad ... ok
test tests::set_version::set_version_good ... ok
test tests::registry_start_app::start_app_nonzero_rc ... FAILED
test tests::set_version::set_version_same ... ok
test tests::upgrade_app::upgrade_good ... ok

failures:

---- tests::registry_start_app::start_app_good stdout ----
[...]

PR #554, opened 2021-02-27
Move shebang line to start of raw string literal

minecraft2048:
Fixes https://github.com/kubos/kubos/issues/550


Ropes:
Thank you for the issue and PR, our recent refresh has resolved the issue by using a format!(...) call.

I ran through your #550 steps, and have not been able to reproduce on `master`.

PR #561, opened 2021-07-31
Fix ci

MoralCode:


MoralCode:
ok github, you really need to stop auto-targeting the repository that i forked from.

im trying to use my own fork to test in production here because i dont know any better way to make sure the changes to the CI work first

PR #585, opened 2022-11-02
Dusting off KubOS

marshall:
this PR makes it possible to build the `kubos-dev` container and KubOS services/apps in both Rust and Python using recent editions of the respective runtimes, namely:
* Python 3.10
* Rust 1.64.0

The biggest change is that all python based code now uses Poetry for dependency management, making standalone python service development much easier to get started with out-of-tree.

Along with this, the `kubos-dev` container has been bumped to 22.04 LTS, and will need to be pushed to the `kubos` docker hub org before this is merged.


Ropes (commented):
Well, I'd love to say all go for merge, but I'm not actually able to validate the changes at the moment. My build vm running Ubuntu 20.20 is being stifled by this perplexing decision by Poetry devs: https://github.com/python-poetry/poetry/issues/6013

They've effectively decided to ignore anyone utilizing `system-site-packages` and declared the issue "won't fix". So I guess it's clear the Python community is keeping up appearances of grand-standing standers over backwards compatibility about standards. I'll have another stab at trying to find a solution, but so far I'm not impressed by the `poetry` devs based on their decisions to provide no workaround for a slightly older systems. Especially if we want to provide long term stability for KubOS.

marshall:
after much pain, all tests are green again, and I think this is ready to merge!

final thing to do is to push the new docker image up to the `kubos` org, and remove my temporary commit bec3c49

plauche:
I'm delighted to see this, as I consider pulling a few pieces of KubOS into a current project 🚀

marshall:
@plauche you lovely human!

marshall:
docker image uploaded and commit reverted

services/scheduler-service/tests/import.rs
```diff
@@ -16,7 +16,7 @@
 
 mod util;
 
-use chrono::Utc;
+use chrono::NaiveDateTime;
```
Ropes, line 19: *hello darkness my old friend*

(change is fine; I had to learn the hard way the differences between the two in a different project)

marshall:
I apparently missed one minor change in the docker image when I pushed, and that has now been addressed along with code review

Ropes (approved):
Let's merge this giant step forward in updates!

PR #591, opened 2022-12-19
Monitor Service loadavg

Ropes:
The ``loadAvg`` query can be used to get information about the system's processor utilization. It operates by reading `/proc/loadavg` and parsing available values to return averaged processor load, counts of active processes, and last process ID status.

Screen Shot 2022-12-19 at 1 37 50 PM

Open to suggestions on naming and pattern. I primarily followed the MemInfo monitor as example.

### DIFF
--- a/docs/ecosystem/services/monitor-service.rst
+++ b/docs/ecosystem/services/monitor-service.rst
@@ -126,6 +126,37 @@ The query has the following response fields:
       without swapping
     - ``lowFree`` - The amount of free memory which can be used by the kernel
 
+LoadAvg Query
+-------------
+
+The ``loadAvg`` query can be used to get information about the system's processor utilization. It
+operates by reading `/proc/loadavg` and parsing available values to return averaged processor
+load, counts of active processes, and last process ID status.
+
+All values are recorded in double precision floats to avoid any potential truncation, and simplify output.
+
+It has the following schema::
+
+    {
+        loadAvg {
+            load1m: Float
+            load5m: Float
+            load15m: Float
+            processesActive: Float
+            processesTotal: Float
+            lastPid: Float
+        }
+    }
+
+The response fields:
+
+    - ``load1m`` - System load averaged over the past minute
+    - ``load5m`` - System load averaged over the past five minutes
+    - ``load15m`` - System load averaged over the past fifteen minutes 
+    - ``processesActive`` - Processes active at the instant of reading /proc/loadavg 
+    - ``processesTotal`` - Processes running on the system in total
+    - ``lastPid`` - Last spawned Process ID on the system
+
 .. note::
 
     Not all response fields are available on all systems.
--- a/services/monitor-service/src/loadavg.rs
+++ b/services/monitor-service/src/loadavg.rs
@@ -32,7 +32,7 @@ impl LoadAvg {
         Default::default()
     }
 
-    fn parse_f64(avg: &str) -> Option<f64> {
+    fn parse_avg(avg: &str) -> Option<f64> {
         avg.parse().ok()
     }
 
@@ -52,9 +52,9 @@ impl LoadAvg {
 
         // Parsers which can safely fail
         if split_vec.len() >= 3 {
-            load_avg.load_1m = Self::parse_f64(split_vec[0]);
-            load_avg.load_5m = Self::parse_f64(split_vec[1]);
-            load_avg.load_15m = Self::parse_f64(split_vec[2]);
+            load_avg.load_1m = Self::parse_avg(split_vec[0]);
+            load_avg.load_5m = Self::parse_avg(split_vec[1]);
+            load_avg.load_15m = Self::parse_avg(split_vec[2]);
         }
 
         if split_vec.len() > 3 {
@@ -123,7 +123,7 @@ mod tests {
 
     #[test]
     fn loadavg_parse() {
-        let info: Result<LoadAvg, failure::Error> = LoadAvg::parse(RAW0);
+        let info = LoadAvg::parse(RAW0);
         assert_eq!(
             info.ok(),
             Some(LoadAvg {
@@ -139,7 +139,7 @@ mod tests {
 
     #[test]
     fn loadavg_partial() {
-        let info: Result<LoadAvg, failure::Error> = LoadAvg::parse(RAW_PARTIAL);
+        let info = LoadAvg::parse(RAW_PARTIAL);
         assert_eq!(
             info.ok(),
             Some(LoadAvg {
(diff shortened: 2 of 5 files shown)


marshall:
love this

docs/ecosystem/services/monitor-service.rst
```diff
@@ -126,6 +126,37 @@ The query has the following response fields:
...
+
+The response fields:
+
+    - ``load1m`` - System load averaged over the past minute
+    - ``load5m`` - System load averaged over the past five minutes
+    - ``load15m`` - System load averaged over the past fifteen minutes 
+    - ``processesActive`` - Processes active at the instant of reading /proc/loadavg 
+    - ``processesTotal`` - Processes running on the system in total
+    - ``lastPid`` - Last spawned Process ID on the system
+
```
marshall, line 159: great docs, thank you!

services/monitor-service/src/loadavg.rs
```diff
@@ -0,0 +1,216 @@
...
+impl LoadAvg {
+    pub fn new() -> Self {
+        Self {
+            load_1m: None,
+            load_5m: None,
+            load_15m: None,
+            processes_active: None,
+            processes_total: None,
+            last_pid: None,
+        }
```
marshall, line 39: just a minor suggestion -- since you already derive `Default`, and the implementation provides None for all `Option` types, you can use it to save quite a bit of work here..
```suggestion
        Default::default()
```

services/monitor-service/src/loadavg.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    fn parse_f64(avg: &str) -> Option<f64> {
+        match avg.parse::<f64>() {
+            Ok(v) => Some(v),
+            Err(_) => None,
+        }
```
marshall, line 46:
```suggestion
        avg.parse().ok()
```

services/monitor-service/src/loadavg.rs
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    fn parse_u64(s: &str) -> Option<u64> {
+        match s.parse::<u64>() {
+            Ok(v) => Some(v),
+            Err(_) => None,
+        }
```
marshall, line 53:
```suggestion
        s.parse().ok()
```

marshall: similar to above, you may not actually want/need these helpers :)

services/monitor-service/src/loadavg.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    // Last process ID launched on the system
+    pub fn last_pid(&self) -> Option<u64> {
+        self.last_pid
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unreadable_literal)]
+mod tests {
```
marshall, line 128: \o/ tests

services/monitor-service/src/loadavg.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    const RAW0: &[u8] = b"0.03 0.05 0.06 1/541 6275\n";
+    const RAW_PARTIAL: &[u8] = b"0.19 0.13 2.6 \n";
+    const RAW_BAD: &[u8] = b"bad data 12.0\n";
+    const RAW_WEIRD: &[u8] = b"0.19 0.13 2.6 bobloblaw 11348\n";
+    const EMPTY: &[u8] = b"   ";
+
+    #[test]
+    fn loadavg_parse() {
+        let info: Result<LoadAvg, failure::Error> = LoadAvg::parse(RAW0);
```
marshall, line 139: FYI, I think this explicit type here is unnecessary.

services/monitor-service/src/objects.rs
```diff
@@ -13,10 +13,36 @@
...
+    field processes_active() -> Option<f64> {
+        self.avgs.processes_active().map(|v| v as f64)
+    }
+    field processes_total() -> Option<f64> {
+        self.avgs.processes_total().map(|v| v as f64)
+    }
+    field last_pid() -> Option<f64> {
+        self.avgs.last_pid().map(|v| v as f64)
+    }
+});
```
marshall, line 44: I know you're following the example of other code in this file, but my radar is up -- it feels like this map conversion is unnecessary. does your code still work if you simply return the raw `Option<f64>` without mapping? it seems like it should

services/monitor-service/src/schema.rs
```diff
@@ -47,6 +48,12 @@ graphql_object!(QueryRoot: Context as "Query" |&self| {
 
         Ok(pids_vec.into_iter().map(PSResponse::new).collect())
     }
+
+    field load_avg(&executor) -> FieldResult<LoadAvgResponse> {
+        loadavg::LoadAvg::from_proc()
+            .map(|avgs| LoadAvgResponse{ avgs })
```
marshall, line 54: nit: space after `LoadAvgResponse`. I think `cargo fmt` should've caught that though hm

marshall (changes requested):
looks great! I left a few suggestions that will make your code a bit more compact and some rust pointers too :)

my only real concern is what looks like unnecessary calls to `map` all over `monitor-service/src/objects.rs`, that aren't necessarily your fault but I'm wondering why the pattern even exists. lmk what you think the best next step is there..

services/monitor-service/src/objects.rs
marshall: I also acknowledge that this is wrapped into a somewhat magic macro, which could be using these map statements as type declarations.. so wasn't 100% certain about it

services/monitor-service/src/loadavg.rs
Ropes: That is some fantastic magic macros!

services/monitor-service/src/loadavg.rs
Ropes: SO much Rust magic to learn.

services/monitor-service/src/schema.rs
Ropes: clippy is sleeping on the job!

marshall (dismissed):

services/monitor-service/src/loadavg.rs
Ropes: good catch!

services/monitor-service/src/objects.rs
Ropes: You're right, I copy-pasta'd then modified this somewhat confusing object conversion. The `f64` values can just return the Option<f64> without the extra map()!

I tried to play with different options of unpacking the u64, then converting without using the map(), but it ended up worse I think, this was the cleanest other implementation I came up with(which is messier IMO):

```rust
Some(self.avgs.processes_active().unwrap_or_else(|| 0) as f64)
```

Given that within the `loadavg` module, keeping the struct fields as integers makes more sense, I think it's fine to convert them here with a map() since GraphQL is the one forcing it be a float?

marshall (approved):
looks great!

services/monitor-service/src/objects.rs
marshall: yeah, good catch -- I didn't realize some of the fields were `u64` :)

in that case, the map makes total sense!

PR #592, opened 2022-12-23
add support for cross-rs compatible aarch64 builds

marshall:
[cross][] seems to be the new all-in-one cross compile front end for Cargo.. it supports Docker out of the box and has some really nice default images. I found setting up `aarch64` builds a breeze, and wasted hours trying to set it up with our current container `kubos/kubos-dev` without success.

I'm wondering if we should move fully to cross for our custom toolchains eventually too.. or just get rid of them. hmmm food for thought :)

basic usage:

1. install [cross][]
1. `cross build --target=aarch64-unknown-linux-gnu`
2. see all the aarch64 binaries under `target`!

[cross]: https://github.com/cross-rs/cross#installation

### DIFF
--- a/tools/dist/Dockerfile
+++ b/tools/dist/Dockerfile
@@ -1,23 +1,33 @@
-# kubos/kubos-dev:1.22.0
+# kubos/kubos-dev:1.23.0
 
 FROM ubuntu:22.04
 
 MAINTAINER marshall@xplore.com
 
 RUN apt-get update -y
+RUN apt-get install --no-install-recommends -y \
+    software-properties-common
 
-RUN apt-get upgrade --no-install-recommends -y \
+RUN add-apt-repository 'deb http://archive.ubuntu.com/ubuntu xenial main universe' && apt-key update
+
+RUN apt-get update -y
+RUN apt-get install --no-install-recommends -y \
     bc \
     build-essential \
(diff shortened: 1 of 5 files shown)


tools/dist/Dockerfile
```diff
@@ -6,18 +6,30 @@ MAINTAINER marshall@xplore.com
...
+    file \
+    git \
+    pkg-config \
+    python3 \
+    python3-pip \
+    python3-setuptools \
+    rsync \
+    sqlite3 \
+    ssh \
+    unzip \
```
Ropes, line 27: This is definitely a nicer grouping of all the build tools!

Ropes (commented):
This looks like a great option to have! 

My only caveat is it'd be nice to have the documentation from the commit and PR somewhere in the docs or a Makefile so that the `cross...` commands aren't lost to the ether. Unless you were thinking it's only for CI?

marshall:
> My only caveat is it'd be nice to have the documentation from the commit and PR somewhere in the docs or a Makefile so that the `cross...` commands aren't lost to the ether. Unless you were thinking it's only for CI?

I should at the very least add this to CI. the primary use case for this is to build the service binaries, which should be a part of the release IMO. good suggestion!

